Issue #202, opened 2015-01-20
SAML Support


pitbulk:
@alexweissman Are you interested in adding [SAML](http://en.wikipedia.org/wiki/Security_Assertion_Markup_Language) support?

I already implemented it in my local environment and support SSO and SLO, I can create a PR if interested.

alexweissman:
Whoa, that's cool!  I was hoping that we'd get some kind of SSO down the road.  

I was originally thinking OAuth, but I don't really know what would be the best for UF.  Can you make a good case for SAML over OAuth?

pitbulk:
SAML and OAuth2 are different protocols.

With SAML you will let your users to log in the system using an external Identity Provider (Salesforce, Google, OneLogin, Okta, ...).
Also there is an open source software named simpleSAMLphp that let you deploy an Identity Provider. With this software you can configure a lot of social authentication, so adding SAML support to UserFrosting you will be able to autenticate users using any Social account (Facebook, Twitter, Google, ....), or any external database/ldap.

alexweissman:
Whoa, I didn't realize that it would be such a simple drop-in!  I was just reading [this article](http://www.mutuallyhuman.com/blog/2013/05/09/choosing-an-sso-strategy-saml-vs-oauth2/):
> SAML has one feature that OAuth2 lacks: the SAML token contains the user identity information (because of signing). With OAuth2, you don't get that out of the box, and instead, the Resource Server needs to make an additional round trip to validate the token with the Authorization Server.

> On the other hand, with OAuth2 you can invalidate an access token on the Authorization Server, and disable it from further access to the Resource Server.

I'm not really sure what this would mean for UF, but it seems like something that can easily be turned on/off by developers.  I'll play with it in a little bit and then probably pull it!

Also, send me an email through my website.  I have some ideas for other authorization features in UF, and it looks like you have quite a bit of experience in that domain.  Would be cool to chat!

pitbulk:
I sent you my mail

ratkobucic:
@pitbulk 
> Are you interested in adding SAML support?
> I already implemented it in my local environment and support SSO and SLO

Hi, I'm trying to add some nice user registration form for one project I work on, and SAML support for UserFrosting would be exact what I'm looking for... :)
So, if you don't mind, can you share your setup/patch for UserFrosting SAML integration?
Thanks +1:

dudheit314:
Any news on this one? I'd really like to use it too!

[...]

PR #204, opened 2015-01-20
Add initial SAML support

pitbulk:
Related to #202
This is a initial approach, as you can see only 2 files of the core were changed (I tried to create it as an external plugin), maybe could be great if we activate the SAML SSO based on some global settings.


alexweissman:
@pitbulk I'm going to close this since the new version is coming out, but I still like the concept.  We should talk about what it would take to implement this as a plugin.

PR #229, opened 2015-02-09
Update bootstrap.css

josh1600:
It makes the Input form look smarter.

See:
http://opensourcecoin.org/aweissman/register.php
before and after adding a width of 39px to the .input-group-addon class.

:)

### DIFF
--- a/css/bootstrap.css
+++ b/css/bootstrap.css
@@ -3969,6 +3969,7 @@ textarea.input-group-sm > .input-group-btn > .btn {
   background-color: #eeeeee;
   border: 1px solid #cccccc;
   border-radius: 4px;
+  width: 39px;
 }
 
 .input-group-addon.input-sm {


alexweissman:
Thanks for the contribution!  Two problems, though:

1. We should try to override the default Bootstrap behavior in `bootstrap-custom.css`, rather than directly modifying `bootstrap.css`.  This will make it easier when we upgrade to newer versions of Bootstrap.  I would also suggest making this a separate CSS class, that way we can pick and choose which forms to apply it to.

2. We should use an `@media` directive so that the style only applies for screen widths 768px and up.  If you try shrinking the window down to mobile sizes, it doesn't look as great ;-)

josh1600:
I had the same concern with adapting the bootstrap file... But just thought u might like to make the form widths for the icon the same as is displays all slightly uneven on my screen.

I found your site today and plan to try out your user management system as I need to build a college project from scratch and I've been told I can use bootstrap and plugins ;)

Keep up the good work!
Josh Thomson - CEO AmazeWebs

alexweissman:
Ohhh, you're trying to make the *icons* the same width.  For some reason when I apply your change, it makes the form fields themselves narrower.

Icons should actually be the same width already - are you using the `fa-fw` class in your icons?

alexweissman:
Fixed icon widths with latest commit.

josh1600:
I saw, nice work :)

I will be using userfrosting for my college project :) ...I will let you know how I get on!

Kind Regards,
Josh.


This email and any files transmitted 
        with it are confidential and intended solely for the use of the individual(s) 
        or entity(ies) to whom they are addressed. If you have received this email 
        in error please notify the system manager. This message contains confidential 
        information and is intended only for the individual(s) named. If you are 
        not the named addressee you should not disseminate, distribute or copy 
        this e-mail. Please notify the sender immediately by e-mail if you have 
        received this e-mail by mistake and delete this e-mail from your system. 
        If you are not the intended recipient(s) you are notified that disclosing, 
        copying, distributing or taking any action in reliance on the contents 
        of this information is strictly prohibited. Data Protection Act 1998: Section 10.


Date: Thu, 12 Feb 2015 11:10:54 -0800

PR #244, opened 2015-02-25
added German translation

rocketworm:
since this is my first pull request, I hope I'm doing this right.


alexweissman:
Yup, you did it right.  Danke!

PR #252, opened 2015-03-03
Added functionality to renderMenu() to allow for menu highlighting of su...

ahouston:
Added functionality to renderMenu() to allow for highlighting of left-sub menus

You can now call renderMenu("site-settings-holder groups") for example in the account/groups.php to open the dropdown and highlight the "Groups" entry.

Currently the parent left_sub entry has no class_name by default, so this needs to be added to the entry (in this example I used "site-settings-holder").


alexweissman:
Sweet, thanks!

Issue #29, opened 2014-05-13
replace all instances of `and` with `&&`


r3wt:
`and` is the incorrect comparison operator and should be avoided entirely as it has lower precendence to `=,>,<,&&,||` etc. 

all instances of `and` should be replaced with `&&` while all instances of `or` should be replaced with double pipe `||`

this can create problems when using and to create an expression between boolean values. 

consider
http://stackoverflow.com/questions/2803321/and-vs-as-operator

PR #253, opened 2015-03-04
Replaced and with &&

dwrj87:
Replaced instances of and with &&, according to #29

### DIFF
--- a/forms/form_user.php
+++ b/forms/form_user.php
@@ -348,13 +348,13 @@ if ($fields['groups']['display'] != "hidden"){
               $group_name
               <span class='pull-right'>
               <input name='select_groups' type='checkbox' class='form-control' data-id='$id' $disable_str";
-          if ((!$populate_fields and $is_default >= 1) || ($populate_fields && isset($user_groups[$id]))){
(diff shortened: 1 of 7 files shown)


alexweissman:
Thanks!  I should let you (and anyone else who is interested in contributing) know that I'm working on a major rewrite of the UF codebase.  So, I wouldn't spend too much time on the current version of the codebase. 

I could, however, really use some help on the rewrite!  If you're interested, hit me up through my website (http://alexanderweissman.com).

alexweissman:
FYI, here is the branch for the rewrite: https://github.com/alexweissman/UserFrosting/tree/slim

alexweissman:
Hi @dwrj87, this patch turned out to have some bugs due to operator precedence: http://php.net/manual/en/language.operators.precedence.php.  Specifically, `&&` has much higher precedence than `and`, so there are a few places where some explicit grouping with parentheses needs to be done.

Would you might fixing that and then resubmitting your pull request?

dwrj87:
@alexweissman I shall take a look at it next week. While I'm at it I shall do the `or`s too

alexweissman:
Thanks!

PR #260, opened 2015-03-11
Remove debug console log for base site path

normington:
Base site path note was turning up in console twice whenever page loaded (presumably the function is called twice when setting up the page).
Removed the console.log that prints site path.

No biggie, but was doing my head in!

Continued to use my installation after deleting this line and nothing exploded.


alexweissman:
Awesome, thanks.

PR #266, opened 2015-03-19
Replace and or

### LINKED ISSUE #29: replace all instances of `and` with `&&`

dwrj87:
Hopefully this bug will fix #29 - unfortunately I can't test it at the moment


alexweissman:
Alright, lemme know if you get a chance to test!

PR #269, opened 2015-03-21
Bug 263 - Lost password email timestamp format fix

anthonybartoli:
This is a fix for https://github.com/alexweissman/UserFrosting/issues/263

$emailDate in models/config.php was formatted incorrectly. 

    $emailDate = date('F j, Y');

Formats a date as MONTH DAY, YEAR (i.e. March 21, 2015)

So the email now reads:

date-fix


alexweissman:
Awesome, thanks!

PR #273, opened 2015-03-22
Documentation of the master branch via ApiGen

anthonybartoli:
See issue https://github.com/alexweissman/UserFrosting/issues/272.


r3wt:
@alexweissman  request for review.

alexweissman:
Let's revisit ApiGen when the new version is ready.  I'm hoping we can completely document the classes in the code itself using codeblocks, then use a tool like this to create a pretty reference.

PR #275, opened 2015-03-22
Refactor - multiple declarations of identical logAllErrors()

anthonybartoli:
There is now one copy of logAllErrors in error_functions.php instead of declaring it several times in different parts of the program. All files that declared logAllErrors already include a copy of error_functions.php.

Also, the logAllErrors() string has been removed from the upgrade instructions since it is no longer declared in config file.

### DIFF
--- a/install/config.php
+++ b/install/config.php
@@ -29,15 +29,6 @@ THE SOFTWARE.
 
 */
 
-// Used to force backend scripts to log errors rather than print them as output
-function logAllErrors($errno, $errstr, $errfile, $errline, array $errcontext) {
-	ini_set("log_errors", 1);
-	ini_set("display_errors", 0);
-	
-    error_log("Error ($errno): $errstr in $errfile on line $errline");
-	throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
-}
-
 require_once("../models/db-settings.php");
 require_once("../models/class_validator.php");
 require_once("../models/password.php");
--- a/models/config.php
+++ b/models/config.php
@@ -29,15 +29,6 @@ THE SOFTWARE.
 
 */
 
-// Used to force backend scripts to log errors rather than print them as output
-function logAllErrors($errno, $errstr, $errfile, $errline, array $errcontext) {
-	ini_set("log_errors", 1);
-	ini_set("display_errors", 0);
-	
-    error_log("Error ($errno): $errstr in $errfile on line $errline");
-	throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
-}
-
 // This will stop the installer / upgrader from running as it normally would and should always be set to false
 // Options TRUE | FALSE bool
 $dev_env = FALSE;
--- a/models/error_functions.php
+++ b/models/error_functions.php
@@ -92,4 +92,13 @@ function checkRequestMode($mode){
 	}
 }
 
+// Used to force backend scripts to log errors rather than print them as output
+function logAllErrors($errno, $errstr, $errfile, $errline, array $errcontext) {
+	ini_set("log_errors", 1);
+	ini_set("display_errors", 0);
+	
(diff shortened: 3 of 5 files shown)


alexweissman:
Nice.  Were you able to test this?

anthonybartoli:
I did not - logAllErrors is passed to set_error_handler in 26 different files. 

It seems to me that the only reasonable and comprehensive way to test this would be to write a unit test for each one of those files and manually trigger an error (perhaps via PHP's built in trigger_error()).

I see that logAllErrors is never directly called - apiReturnError is used often (always?). I don't yet see the connection between apiReturnError and logAllErrors though. I'm still getting up to speed on the codebase.

What are your thoughts on using PHPUnit? I've read that Slim includes them by default, so perhaps it would be easier to incorporate tests in the next version of UF.

alexweissman:
I wouldn't waste too much time delving into the current codebase, especially the error- and control-handling.  Slim takes care of all that in the next version.  `logAllErrors` is just a function to tell UF to convert PHP errors to exceptions, which can then be caught in `try...catch` blocks.  

`apiReturnError` basically halts the script and generates an appropriate response depending on whether the original request was AJAX or not.  But, Slim's error-handling is much more elegant and can do all of this automatically.

Unit testing is a good idea, and I know it is part of the best practices for modern development.  Unfortunately, I don't have that much experience with it.  I've encountered the term "design for testability" in a few blog posts I've read, mainly suggesting that dependency injection, factory patterns, etc can make it easier to write testable code.  Can we still do unit testing even if we don't strictly follow those patterns?

anthonybartoli:
>>Can we still do unit testing even if we don't strictly follow those patterns?

I will find out this week. I'm working through a couple testing books right now.

alexweissman:
Sweet, thanks for looking into this.

anthonybartoli:
>>I've encountered the term "design for testability" in a few blog posts I've read, mainly suggesting that dependency injection, factory patterns, etc can make it easier to write testable code. Can we still do unit testing even if we don't strictly follow those patterns?

I'm fairly certain that we can test any code that is part of a class or a function. It's the code not part of a function or a class that is much harder to test (and perhaps untestable) since it cannot be (easily) contained or called separately from the application in an automated fashion.  

>>logAllErrors is just a function to tell UF to convert PHP errors to exceptions, which can then be caught in try...catch blocks. 

I'm seeing try/catch blocks in:
-install_db.php (not in a function or a class)
-wizard_site_config.php (not in a function or in a class)
-authorization.php (checkActionPermits()) 
-db_functions.php (checkBinaryConfig())
-0.2.2.install.php (not in a function or a class)

Given that I just moved a function into a file that was already being included, it may not be worth the effort to test the above files. I may do so to learn more about testing though.

alexweissman:
The other thing to consider is that Slim has its own [error-catching mechanisms](http://docs.slimframework.com/#Error-Handling-Overview), which I'd like to use as the basis for our error handling system. 

Basically, my plan is to use a custom error-handler in Slim: https://github.com/alexweissman/UserFrosting/blob/slim/userfrosting/config-userfrosting.php#L110

This will do two things when an Exception is raised:

1. Log the exception info to the PHP error log
2. Add a generic "Server error" message to the error log.  That way, we let the client know there was an error but without exposing any details.  Only the developers/server admins will be able to see the errors.

r3wt:
@alexweissman  whats up with this pull? we need to start labeling pulls so contributors can know whats going on.

r3wt:
@alexweissman Please merge or close this PR

alexweissman:
I have implemented exception-handling for 0.3.0: https://github.com/alexweissman/UserFrosting/blob/dev-0.3.0/userfrosting/config-userfrosting.php#L135.  Let's not worry about fixing it in 0.2.7.

PR #282, opened 2015-03-28
added Dutch translation

joey102:
Dutch translation for those from the Netherlands.


alexweissman:
dankejewel!  And, welcome to UF :-)

PR #377, opened 2015-07-25
fix footer overflow

frostbitten:
Setting the footer's container to container-fluid will allow it to be small at a large resolution (such as when the sidebar takes some of the width).


alexweissman:
Thanks!

PR #425, opened 2015-10-18
Fix problem in `MySQLDatabase::connection()` method

### LINKED ISSUE #424: HHVM Compatibility issues
so problem arises in `MySQLDatabase::connection()` method. I think we should change line 29 from `echo` to `die()` as a temporary solution. otherwise the script never terminates until the fatal is thrown and the user never sees the message of db failure, because its captured (and discarded due to the fatal error) by Slim's Output Buffering.

r3wt:
script should be halted here due to lack of db connection. this fixes the problem where it is not, and the error message gets swallowed. temporary workaround until we can get `DatabaseException` to control flow for these situations. closes #424

### DIFF
--- a/nginx_vhost.conf
+++ b/nginx_vhost.conf
@@ -1,0 +1,43 @@
+
+# This file is an example vhost for nginx.
+# you may place this in a seperate file and include it, 
+# or you may place this directly withing the http block of the nginx.conf
+
+server {
+    listen 80;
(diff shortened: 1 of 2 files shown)


r3wt:
closes #424

r3wt:
closing this for now. we can dig it up again later if we need to.

PR #475, opened 2015-12-09
Register to return User Object

ssnukala:
Updated Accountcontroller/register function to return the user object.
This will help the plugins using custom registration utilize the out of
the box registration process.


alexweissman:
Manually merged.  I accepted your changes to `AccountController:register`, but I am rejecting your changes to `userfrosting.js`.  We should come up with a more robust way to allow developers to control the "binarization" of individual checkboxes - please open an issue to discuss.

PR #493, opened 2016-01-12
optional enable twig cache

blink01:
Works great for me. I'm eager to here you feedback.

### DIFF
--- a/userfrosting/config-userfrosting-example.php
+++ b/userfrosting/config-userfrosting-example.php
@@ -26,6 +26,8 @@
             'log.enable' => true,
             'debug' => false,
             'base.path'     => __DIR__,
+            'twig.cache.enable' => false,
+            'twig.cache.path' => __DIR__ . '/cache',
             'templates.path' => __DIR__ . '/templates',     // This will be overridden anyway by the default theme.
             'themes.path'    =>  __DIR__ . '/templates/themes',
             'plugins.path' => __DIR__ . '/plugins',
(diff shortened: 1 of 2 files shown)


Netrilix:
I just confirmed the Twig will create the `cache` folder if it doesn't already exist, so there's no additional configuration required there.

r3wt:
@frostbitten that needs to be reverted IMO. If the dir is not owned by the webserver user, twig will die with a fatal error. 

Checks need to be performed to ensure webserver owns the directory or has permission to write to the directory. if it doesnt we should either:

a) throw exception
b) silently ignore directive and write to error log indicating that cacheing is disabled due to permissions/ownership error.

alexweissman:
Yes, I agree with @r3wt .  Also, any change to the code should also include an update to the changelog and version number in the DB.

Netrilix:
Nice catch @r3wt . I thought of the directory not existing, but not of it not being accessible at all.

r3wt:
No prob, i just happened to be in the neighborhood and recalling an embarrassing production issue from days gone by lol

frostbitten:
Ah, okay. Sorry for that, didn't go over the pr checklist. Haven't looked into how twig does it, but if it uses mkdir shoudln't the created folder be guaranteed to be 0777?

alexweissman:
Yes, but PHP will still need permission for the parent directory.

Netrilix:
We'll also have to look into the scenario where the directory itself *does* exist, but isn't writable.

PR #507, opened 2016-01-22
Add missing lines of Thai language

popiazaza:


r3wt:
looks ok to me.

@alexweissman @Netrilix All in favor of merging?

Netrilix:
I'm in favor

alexweissman:
I can't read Thai, but this looks good to me.  Can we add this in with the next hotfix instead of merging it directly?

Netrilix:
Yeah, I'll add it to the list.

alexweissman:
Great.  This way, we can merge everything into both the master and dev branches.

Netrilix:
This was merged into `hotfix-0.3.1.12`, which should be merged into `master` in the next day or so.

Issue #473, opened 2015-12-07
Status/Action always Active


fv3rdugo:
In version 0.3.1 in the user area you see always Active status although status is Disabled or Unactivated. If you change the status this changes are applied in the database but you can't see.
thanks!

alexweissman:
Is this a fresh install of 0.3.1, or are you upgrading from 0.3.0?

fv3rdugo:
Fresh install of 0.3.1

alexweissman:
So, are you talking about in the "Users" page?

![screen shot 2015-12-07 at 1 00 01 pm](https://cloud.githubusercontent.com/assets/5004534/11635095/7c5d0e66-9ce2-11e5-8cc1-77d53b3dfd40.png)

fv3rdugo:
Yes, in this page.

alexweissman:
What happens if you just view the raw JSON data from the user list API?  Go to `/public/api/users` and show me what you see.

fv3rdugo:
Are the same results that I can see in database.

{
    "count": 3,
    "rows": [
        {
            "id": "4",
            "user_name": "admin",
            "display_name": "admin",
            "email": "xxxxx@gmail.com",
            "title": "New User",
            "locale": "en_US",
            "primary_group_id": "1",
            "flag_verified": "1",
            "flag_enabled": "0",
            "flag_password_reset": "1",
            "created_at": "2015-12-07 13:12:16",
            "updated_at": "2015-12-07 13:13:36",
            "last_sign_in_time": 0,
            "sign_up_time": "2015-12-07 19:12:16"
        },
        {
            "id": "2",
            "user_name": "pepe",
            "display_name": "pepe",
            "email": "pepe@xxxxx.com",
            "title": "New User",
            "locale": "en_US",
            "primary_group_id": "1",
            "flag_verified": "0",
            "flag_enabled": "0",
            "flag_password_reset": "0",
            "created_at": "2015-12-06 17:25:06",
            "updated_at": "2015-12-07 12:08:23",
            "last_sign_in_time": 0,
            "sign_up_time": "2015-12-06 23:25:06"
        },
        {
            "id": "1",
            "user_name": "root",
            "display_name": "master",
            "email": "xxxx@gmail.com",
            "title": "New User",
            "locale": "en_US",
            "primary_group_id": "1",
            "flag_verified": "1",
            "flag_enabled": "1",
            "flag_password_reset": "0",
[...]

PR #515, opened 2016-02-11
Solution to issue #473

Aterbonus:
The problem was the strict comparison.


alexweissman:
Thanks.  Let's merge this in with the next hotfix.

PR #522, opened 2016-03-05
updated text to highlight database table prefixes

leanpanda:
and how PHP code should not include the prefix


alexweissman:
Thanks, I've updated lesson 3 to make this clearer (but in a slightly different way than you did).

PR #564, opened 2016-07-06
update page404() to respond with 404 code

frostbitten:
Takes advantage of twig's ability to render a template to a string and slim's halt() method.


frostbitten:
Looks like I accidentally copied over a hook I have in my setup. Although a nice tool, I'm not sure if we'd want to add the hook.

Netrilix:
What are you using the 404 hook for? I'd imagine we'd probably want to merge without the hook, but we should at least consider it if you think it's useful.

frostbitten:
I had been using the hook to tell another framework to boot (which has its own router). An average use case however would would be to trigger a custom log event.

frostbitten:
Alrighty, hook removed as it's not related at all. I also did the same technique with `pageDatabaseError()`. Does `500` seem to be the [appropriate response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) code? Or should we go with `503` since 

> UserFrosting needs to connect to the database for just about every request.

Netrilix:
I'm a big fan of this chart: http://mlk.nfshost.com/choosing-an-http-status-code/HTTP-5XX-Status-Codes.svg

I'd lean toward 503, since a database server restart is going to throw this error, and the user can try again in a minute or two and be completely fine. There's a good note on your link about caching though:

> This responses should be used for temporary conditions and the Retry-After: HTTP header should, if possible, contain the estimated time before the recovery of the service. The webmaster must also take care about the caching-related headers that are sent along with this response, as these temporary condition responses should usually not be cached.

alexweissman:
This looks like a good clean, temporary solution until v4, when we will [generate HTTP error statuses through the error handler](https://github.com/userfrosting/UserFrosting/blob/dev/app/core/src/Handler/UserFrostingErrorHandler.php).

503 is a more specific error condition, which is good.  I know that 503 is also frequently used for when a server is overloaded - for example, during Reddit's peak hours.  But, I think it is the closest we will be able to find unless we start using custom error codes (which could create their own problems).

frostbitten:
Nice flow chart @Netrilix, very handy. I was googling around and found a [nice stackoverflow](http://stackoverflow.com/questions/942951/rest-api-error-return-good-practices) on this topic with more of those charts, though I can't seem to peg the original author/site.

I pushed the change for a 503 code and added a call for pageDatabaseError() if UserSession hits a PDOException. Seems like a good idea, and not seeing any issues myself when testing.

Netrilix:
@frostbitten I got the chart from here: http://racksburg.com/choosing-an-http-status-code/

@alexweissman 
I just got an error on reddit that looked like this:
>sorry, something broke on our end
please try again in a minute
(error code: 502ssl)

That's saying that an external service (ssl) is responding but not properly. Would that be better, since we've got a specific service we're pointing to that's not working? I'm leaning toward the "try again soon" nature of 503, but I'm not sure.

frostbitten:
From my little understanding of 502 Bad Gateway responses it seems we would need to be acting as a proxy for that error to be worthwhile. I've seen it on cloudflare [when my server is down](http://i.imgur.com/r7c8fu9.png)(not my site) and it makes sense there. Not sure to what extent the term "gateway" covers. Specifically, if a user interacts with the site and an sql server is used, does that mean the user was accessing the sql server through a gateway? This is where I'm not sure

Netrilix:
I think you're right. I saw one very general definition that defined an application gateway as an application that translates from one data form (eg. SQL) to another (eg. HTML), but the rest of the definitions are pretty clearly network related, as you mentioned.

frostbitten:
Okay, then I think we're all set to merge. I'll wait for final word from @alexweissman

Netrilix:
Could you change the PR to point to my `.16` hotfix branch? Then I can merge it into master properly.

frostbitten:
done with #567

PR #567, opened 2016-07-09
update page404() to respond with 404 code and other

frostbitten:
Also, use `503` for pageDatabaseError(). See individual commits for more detail


alexweissman:
Has this been tested with the installer?  You'll notice the comment:

> // If we can't connect to the DB, then we can't create an authenticated user.  That's ok if we're in installation mode.

Issue #560, opened 2016-06-26
Considerations for assets in v4


alexweissman:
There's been a lot of hubbub about three related topics:
### Separating developer and core code

This is the biggest one.  To build a site with UF requires the developer to directly change the code that ships with the initial installation.  This makes it harder to upgrade in the future, even when using `git merge`.  Also, I've found certain applications where I actually want to _remove_ some functionality from UF.  For example, sometimes I create a purely public site, and I need to strip out all of the authentication and account management features.   
### Theming

To me, this means a drop-in package of custom styling and templates.  Themes could be applied to an entire site, or be customizable for different groups/users.  We sort of have this right now, but it is fairly limited in what it can do.  For example, all custom CSS has to go into a single CSS file, and a theme can't incorporate its own client-side code.
### Plugins

A drop-in package of code that implements some functionality.  For example, we have the OAuth and thumbnail plugins.  They rely on an awkward combination of hooks and inheritance to function properly.

---

**What these all have in common**, is that they are all self-contained module that should require no modification of the base code.  My idea is to generalize these three types of modules to be handled by the UF application in the same way, which we are calling _sprinkles_.  Thus, developer-contributed code, third-party plugins, and themes would all be types of sprinkles, and placed in subdirectories under the `/sprinkles` directory.

This brings us to the subject of **assets**, i.e. CSS, Javascript, image files, and other media files.  I've developed the [userfrosting/assets](https://github.com/userfrosting/assets) package, which can generate the appropriate `<style>` and `<script>` tags from a [JSON schema](https://github.com/userfrosting/UserFrosting/blob/dev/build/bundle.config.json).  However, I am not sure if this fully meets the needs of all possible use cases for sprinkles.  Consider the following use cases:
- A sprinkle should be able to contribute assets to site pages - both those pages managed by the sprinkle itself, but also any of the "core" pages.  This is essentially the behavior people associate with "theming".
- A sprinkle should be able to _override_, _replace_, or _remove_ assets from the core or other sprinkles.  For example, we could imagine a sprinkle that replaces the default `bootstrap-3` with `bootstrap-4`.
- In the case of themes, it should be possible to dynamically select certain theme(s) at runtime.  For example, group- or user- specific theming.
- The "effective assets", i.e. those that will actually be active on the site once all overrides/replacements/removals have been resolved, need to be compiled and placed in the `public/` compiled asset directory.  Note that in the case of dynamically selected themes, there will need to be a compiled version of each possible theme's assets. 

[...]

PR #570, opened 2016-07-10
Proposition for plugins improvements and separation of code

lcharette:
Here is something I've been working on over the weekend regarding what's been discussed over in #560 and on the chat about separation of UF core code ans user code. Those changes allowed me to move all the custom code I'm using to build my new website in a dedicated folder inside the existing plugin directory. Like discussed before, the ultimate goal this is to separate (my) code from the UF core code so when it comes time for an update, everything will be easier.

# What did I changed?
## Added hooks at some key point in core UF code to be able to extend some of the basic functionality
Started with some basic ones and more could be added as we work with this. But those are enough for now to make the below changes work. This include when the routes are defined, when the guest user is loaded, when the authenticated user is loaded (so you can add stuff to the user object *wink wink*) and when the whole system is ready.

## Added /routes to plugins
So plugins can define their own routes in their directory instead of doing everything in `index.php`. Also enable to have multiples files for the same project so all the routes can be separated by topic for easier navigation.

## Added template hooks and plugin template discovery
Added hooks to include twig files at specific locations in the template from php. Perfect to add menu items or site wide stuff from php and plugins. More on the template later.

## Translation as folders and plugin locale discovery
This one comes in two parts: First, translations table should be able to come in multiples file format. Having a single file for **every keys** will be a total nightmare pretty quickly on large sites. By having the translations tables in multiples files inside a `fr_FR` directory, the translator jobs become much easier. Second, this is necessary for plugins (or custom code) that needs to bundle their own translation tables. So a little modification was needed on the *Fortress* side.

## Moved demo and placeholder stuff to the example plugin
Because when I update, I don't want to remove that again ;)
For the dashboard, I'm still not sure how to handle it... We don't have a universal homepage to link to...

## Added "Breadcrumbs" demo plugin
Just a little something I use on my website for easier navigation in plugin form for demo purposes.

# Current limitations, things to consider and where to go from now
First, you may ask about controllers, models, schema, etc. Turns out composer was already setup for that. So drop the plugin in, call `composer update` and everything should be working. If you want to try the `Breadcrumbs` plugin I talked about, you'll get an error if you don't do `composer update`.

Some stuff still need to be done:
 - Browser assets (javascript, css, images, etc.) inside plugins. What to do with them since they are not in the public directory?
 - Priority of code : Which plugins come first? Which hook is called/displayed first?
 - Dependencies : When plugins start distributing (if they ever do), we'll need some sort of dependencies checker. For example, since the code for my website call the `Breadcrumbs` plugin, I'll have an error and wsod if this plugin is missing. Composer can probably help.

Like some of us discussed on issue #560 and on the chat, there will probably never be a "perfect" setup on how to handle plugins and separation of UF core code from user code. Users will always be able to edit the core files how they want, but at least this open a door for better way of doing things, easier integrations of updates and new stuff and easier support. There'll never be enough hooks, especially on the template side, but at least if 80% of use case are taken care of, it's gonna be a good start.

Speaking of template, this one is really tricky. When working with CMS or pre-built softwares (a forum platform for example), users are somewhat expected to use the default template or theme (whatever you call it) or a pre-built one found on the Internet. With UF I agree it is completely different. While the dashboard template is perfect for the project I'm currently working on, it probably won't be for someone else. I guess time will tell us what users are doing with UF and how to handle this. Or maybe someone will have a better idea...

Anyway, I don't think the way I handle plugins right now is ready for prime time (maybe I've done something Composer can take care of?), but at least it's something to start working on. The goal of this pull request is not necessary to merge it right away in UF code, but to share what I've been working on (which do work for me) and open discussion for future UF release. Maybe `plugins` is not even the right word for this, who knows... 

# One more things: what it looks like IRL?
For the curious ones, here's my project current structure. Everything outside `plugins/` represent the current status of my UF branch + the config files. So pretty 😄 
uf_cpm_structure


Netrilix:
[Closed automatically when I did an amended commit on the hotfix branch]

PR #585, opened 2016-08-04
New MessageTranslator with Repository style file loading and plural management system for UserFrosting V4

lcharette:
Now more than one files can be loaded per Sprinkle, nested key can be accessed using dot syntax and a new plural system is available. 

The usage of MessageTranslator is the same as before, but some functionality are added. Languages files can now define subkeys for easier navigation and access:
```
"ACCOUNT" => array(
    "SINGUP" => "Sign Up",
    "REGISTER" => "Register"
)
```
Nested keys an be called using dot syntax:  `{{translate("ACCOUNT.SINGUP")}}`

A new plural system allow for easy pluralization of strings. This whole system is based on Mozilla and phpBB works (See : https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals). Languages files are defined like this, using the plural rules with the `zero` special case:
```
"NEW_MESSAGE" => array(
	0 => "No new message",
	1 => "You have one new message",
	2 => "You have {{int}} new messages",
)
```

The correct plural form will be available in the `int` placeholder using `{{translate("NEW_MESSAGE", {int: 5})}}` or the shortcut `{{translate("NEW_MESSAGE", 5)}}` if you don't have others keys. The default `int` key can be overwritten using a third new option like this: `{{translate("NEW_MESSAGE", {nb: 5}, 'nb')}}`. This may be useful if you pass an existing array to the translate function.

If a localized string contain more than more plural, for example `1 guest and 4 friends currently online`, the easiest way to apply the plural rule to both `guest` and `friends` would be to nest translation functions:
```
$online_guest => translate("ONLINE_GUEST", [guest => 1]);
$online_friend => translate("ONLINE_FRIEND", [guest => 1]);
translate("ONLINE_USERS", ["guest" => online_guest, "friend" => $online_friend]);
```

Where `ONLINE_USERS => "{{guest}} and {{friend}} currently online";`

Those example using the Twig syntax can also be used in php.


alexweissman:
Ok, let's wait till we've got the pluralization API finalized before we start using the new version of i18n in dev.

lcharette:
Everything should be ok now and clean for PR merge. Removed the file in vendor as the one that will be fetched from `i18n` dev will be the last version. It shows me that the branch has conflicts, but I can't see what are those conflicts. Let me know if it's something I can fix, otherwise I could delete this PR and open a new one from a fresh fork.

lcharette:
Nevermind, fixed the conflict. Good to go now!

lcharette:
This is outdated. See #590 and `i18n`

Issue #541, opened 2016-05-02
Use Illuminate/Schema in installer, instead of raw queries, to support SQL Server, Postgres, and SQLite automatically


xoorath:
Hey,

I'm investigating userfrosting for an upcoming project. We would like to use our azure SQL db but I'm not sure how to connect successfully from userfrosting.

A connection would normally be made with the following:
`$connection = new PDO( "sqlsrv:server = tcp:$HOST,$PORT; Database = $DB", $USER, $PASS);`

But I'm unsure how I can tell userfrosting to make a connection with sqlsrv, or include tcp and the port in the hostname? I guess for the tcp+port I would just set `db_host` to include tcp and the port? but I'm still not sure how I could configure userfrosting to use sqlsrv...

Any ideas?

alexweissman:
Eloquent (the database abstraction package we use) definitely supports SQL Server.  The connection is initialized in [initialize.php](https://github.com/userfrosting/UserFrosting/blob/master/userfrosting/initialize.php#L42-L51).  

For more information on `Capsule`, which is the standalone adapter for Eloquent, see https://github.com/illuminate/database/blob/master/README.md.

xoorath:
Hey @alexweissman,

I'm sorry to bug you about this, I'm sure this isn't a userfrosting problem. If you could help point me in the right direction that would be really appreciated though....

I think I must be writing my connection strings wrong, I can't connect using any variation other than _exactly_ the string I posted above, that I got from azure. I'm not sure how to format my connection parameters in userfrosting so I end up with the "sqlsrv: server=tcp:host"<and so on>. 

Do you think I might be missing some configuration step in azure, or if it's a config issue? Any advice is appreciated.... Thanks so much.

xoorath:
Hey,

So I'm making some progress. I added a `db_port` option to the config and linked it into the `initialize.php`, and I also changed the driver in the `initializephp` to `sqlsrv` like our azure database is expecting. So at least I can connect now.

However now we're failing when trying to execute a query. It's complaining that the SQL isn't valid, probably because sql server expects TSQL.

The error I get is (with trimmed paths and comments removed for brevity):
`
Error in vendor\\illuminate\\database\\Connection.php on line 624: SQLSTATE[42000]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]Incorrect syntax near the keyword 'IF'. (SQL: CREATE TABLE IF NOT EXISTS `uf_configuration` (
[...]

PR #599, opened 2016-11-24
Azure SQL Server and SQL Server Support

Silic0nS0ldier:
Azure SQL Server and SQL Server Support as discussed in  https://github.com/userfrosting/UserFrosting/issues/541.

Tested with Azure SQL Server (site hosted in localhost, database at Microsoft Azure) and MySQL/MariaDB in localhost.

### DIFF
--- a/userfrosting/controllers/InstallController.php
+++ b/userfrosting/controllers/InstallController.php
@@ -230,6 +230,7 @@ class InstallController extends \UserFrosting\BaseController {
 
         // Create the master user
         $user = new User($data);
+        
         $user->id = $this->_app->config('user_id_master');
 
         // Add user to default groups, including default primary group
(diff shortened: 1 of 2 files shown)


alexweissman:
Looking good - but what's the issue in `InstallController`?  Why was that `save()` giving you trouble?  It doesn't seem to set the `id` explicitly anywhere.

Silic0nS0ldier:
See line 234 of the install controller.

alexweissman:
Oh!  So, trying to set the `id` through Eloquent's active record is giving you the same problem?

I assume the issue is with:

`$user->id = $this->_app->config('user_id_master');`

Silic0nS0ldier:
Yep. Hence me putting in the raw T-SQL statements for when its Sql Server. I tried enabling the id specification before it, but no luck as is server considered them to be different sessions (and the necessary setting resets to default once the session ends).

alexweissman:
Gotcha.  This is fine as a workaround for now, but we should use parameter binding instead of interpolating variables in the query, since this opens you up to SQL injection.

Silic0nS0ldier:
You mean like described at http://www.w3schools.com/php/php_mysql_prepared_statements.asp ?

alexweissman:
Yep!  But in Laravel, it is exposed via the `setBindings` method.  See http://stackoverflow.com/a/20873009/2970321

Actually, you probably don't even need that level of control - why not just use `insert`?  Then you could convert the user data to an array using `toArray()` and just pass it into the call to `insert`.

Silic0nS0ldier:
Took a ludicrous amount of googling and testing, but I've got it working.

Something I did stumble upon something else though, it seems that the JS and CSS minification is busted.

alexweissman:
So, maybe a silly question, but...couldn't you just call the `SET IDENTITY_INSERT [$tableName] ON;` query, and then use `->save()` like normal, instead of having to add that explicit insert query afterwards?

Silic0nS0ldier:
If only. The trouble is that it would be considered a different session, and as such SQL Server resets the `IDENTITY_INSERT` variable to default before you can call `->save()`. In the end, this was the cleanest I could get it.

But who knows? Maybe the people who work on the ORM library will do something nice in the future, like providing a way to allow easy specification of an identity in SQL Server. Probably not though, given that PHP + SQL Server is a rare combo.

alexweissman:
What if you tried using a [database transaction](https://laravel.com/docs/5.2/database#database-transactions)?  That might keep the session open...

Silic0nS0ldier:
Did try transactions (if you look at the first commit, you'll see that I forgot to remove one after attempting that approach).

All transaction does is make it easy to rollback changes. Each interaction with the database is still treated as its own session.

alexweissman:
Weird.  Ok, I'll test this and hopefully get it pulled in!

alexweissman:
Ok, a couple of things:

1. You should use Carbon for the date (`\Carbon\Carbon::now();`) instead of PHP's native `date` function.
2. In your custom logic block for SQL server, the root user doesn't get added to the default groups.  I think this is because you reload the `$user` object after the groups have been added.

Silic0nS0ldier:
Not sure how I missed the default groups. Nevertheless, it should all be good now.

lcharette:
Looks like this introduced a bug. See : https://github.com/userfrosting/UserFrosting/commit/5f3cb538ed4086f9afc71758c2e2ee53d533f7e3#commitcomment-20142743

PR #603, opened 2016-12-11
MySQL Zero Date Fix

Silic0nS0ldier:
Made `timestamps` nullable to prevent conflict with MySQL modes 'NO_ZERO_IN_DATE' and 'NO_ZERO_DATE'.

Regular `timestamp` does not conflict with the 2 modes, as its default value is `CURRENT_TIMESTAMP` (or equivalent for the used database system).

Tested with Azure SQL Server, MariaDB (via XAMPP), and MySQL (via WAMP).


lcharette:
If you tested it with both db... sounds good ! Thanks !

alexweissman:
Merged.  Thanks!  https://github.com/userfrosting/UserFrosting/commit/89101e4e96670ae305dd85be0ba9c6c462dedb2e

PR #639, opened 2017-02-06
Add translation for Portuguese (locale pt_PT)

brunomnsilva:
Add the translation strings for core/admin/account sprinkles.


lcharette:
Nice! Can you add it to the list of available locale here: 
https://github.com/userfrosting/UserFrosting/blob/develop/app/sprinkles/core/config/default.php#L130-L133

brunomnsilva:
I think its there now!

PR #651, opened 2017-02-17
Add Thai translation.

popiazaza:
Thai translation for UF4 is here. Cheers!


alexweissman:
Whoa, that was so fast!  Thanks!

PR #673, opened 2017-03-21
Prevent mixed content on demo pages

splitt3r:
Load images from placehold.it with the protocoll used to access the page.


alexweissman:
Thanks, we'll merge this into the next release!

app/sprinkles/core/assets/local/core/js/uf-alerts.js
```diff
@@ -173,13 +173,13 @@
      * Push a given message to the current uf-alerts collection.
      *
      */
-    Plugin.prototype.push = function (type, message)
+    Plugin.prototype.push = function (options)
```
lcharette: Oh ok. See : https://github.com/splitt3r/UserFrosting/blob/1ce354aed07410f059f716182ec31002f9965662/app/sprinkles/core/assets/local/core/js/uf-alerts.js#L301

When using `push` to manually push an alert inside javascript (i.e. `$("#alerts-page").ufAlerts('clear').ufAlerts('push', 'danger', 'Hello World').ufAlerts('render')` ), it would fail since the above linked line would pass the arguments as an array to the `push` method. Push doesn't seems to be used anywhere else on the system, so no breaking changes.

PR #676, opened 2017-03-23
Update install.php

x00x70:
make 'Y' the default behaviour for STDIN

### DIFF
--- a/migrations/install.php
+++ b/migrations/install.php
@@ -74,6 +74,7 @@
     echo "Is this correct?  [Y/n]: ";
 
     $answer = trim(fgets(STDIN));
+    if(empty($answer)) { $answer = "Y" ; }
 
     if (!in_array(strtolower($answer), array('yes', 'y'))) {
         // OS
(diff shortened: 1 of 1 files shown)


Silic0nS0ldier:
LGTM, but it could use some way to indicate the default action.
Especially since this will invert behaviour.

x00x70:
Yeah - i thought that as 'Y' was uppercase this indicates it is the default (as per most unix programmes). see - http://askubuntu.com/a/322601

Silic0nS0ldier:
Windows background. :wink: PowerShell is much more verbose with Yes/No scenarios, and command prompt rarely has a Yes/No scenario.

Either way, good to know.

Silic0nS0ldier:
(personally though, I quite like the `([y]/n)` version, if only because its hard to miss)

x00x70:
done.

Silic0nS0ldier:
Cheers. Should make the behaviour much more apparent to the user base (and prevent me from making mistakes when I'm tired).

PR #677, opened 2017-03-24
Update AccountController.php

x00x70:
Changed default behaviour to redirect to index instead of admin (dashboard).


lcharette:
+1 on this. Some app might not have a public dashboard for regular (non-admin) users. Also, I think it's gonna be easier for dev to change the `/` route than the `admin` one (most sites won't use the default index controller anyway).

Plus, it's not a redirect that will be used often...

splitt3r:
The comment should be changed too :smile:

lcharette:
Since the solution might be a little more complicated, I opened issue #680 and I'll close this PR for now.

PR #684, opened 2017-04-05
Added scripts to build and run the docker container

dimm0:
This way of building the container worked well for me


alexweissman:
Please see #696, which breaks out each service (php, mysql, nginx) into a separate Docker container.

PR #685, opened 2017-04-06
User must verify account again after changing email

splitt3r:
If the `site.registration.require_email_verification` config value is `true` the user must validate the account again after changing the email address, This prevents users from entering fake emails after the account is verified once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # Ignores my text editor metadata
 *.komodoproject
+*.iml
+.idea/
 
 # Ignores Mac metadata.  You can configure this globally if you use a Mac: http://islegend.com/development/setting-global-gitignore-mac-windows/
 .DS_Store
@@ -20,7 +22,7 @@ app/logs/*
 app/sessions/*
 
 # Ignore bower vendor assets
-/app/sprinkles/*/assets/vendor/
+app/sprinkles/*/assets/vendor/
 
 # Ignore sprinkles.json config file
 app/sprinkles.json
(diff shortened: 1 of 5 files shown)


alexweissman:
So, what's the workflow like with this?  When they change their email address, they immediately get logged out and then have to reverify the address before they can sign back in?  I'm not sure that would result in the best user experience.

splitt3r:
Yeah perhaps to aggressive. I will try to figure out a better way to do this.

alexweissman:
I have an idea.

alexweissman:
I've seen a lot of sites that have a "primary email address", and then one or more secondary email addresses.  We could do something like this:

- primary email address is stored in the `users` table;
- secondary email addresses are stored in a separate table that also keeps track of their verification statuses;
- primary email address must be a verified email address;
- you must always have a primary address;
- the primary email address can be changed, but only to a secondary email address that has already been verified.

Still, I'm not sure if this is necessarily something we need to implement in core.  It would make a really good community Sprinkle for sure.

splitt3r:
The Idea may be a good community sprinkle but i think the main thing is an important bug. UF has the E-Mail login Option but i can change my email to an unverified one. So if i know wich E-Mail you are using for login that could make problems.

lcharette:
Could your fix be activated by a setting?

I do prefer Alex solution, but as something built in. GitHub uses a similar method. 

or could be a mix of both. You change your email, but the old one is not replaced with the new one (old one get removed) until the new one is verified.

splitt3r:
I think saving the old email until the new one is verified is a good way to go :) perhaps we could use the throttle data field to save the new one and update the profile field afer the verification succeeds. I will have a look at it after the weekend.

alexweissman:
Incidentally, this is a useful article about the rationale behind requiring email verification in the first place: http://softwareengineering.stackexchange.com/questions/232562/why-do-most-sites-require-email-activation

alexweissman:
@splitt3r can you open a new issue, and summarize the relevant details of our proposed solutions, so we don't lose track of this?

splitt3r:
yep

PR #687, opened 2017-04-07
Update sign-in.js

x00x70:
Removes leading slashes from redirect string and also excess trailing slashes to make URIs look pretty :)


alexweissman:
Looks great, thanks!

PR #691, opened 2017-04-12
Hotfix for ufTable

splitt3r:
The attribute `data-sortlist="[[0, 1]]"` should indicate wich column should be used to sort the table by default. But indeed it doesn´t as you can see in the url `../admin/activities#&sort[table-activities][occurred_at]=desc&page[table-activities]=1&size[table-activities]=10` because `occurred_at` is the first column and not the second. This PR fixes that.


lcharette:
So... I figure out this fix is not needed?

splitt3r:
Created the pr to fast :D this does not correctly fix the problem.

lcharette:
The table sorter has some weird caching going on. Try clearing the table sorter local storage from the browser console and then reload the page. You'll be sure to have the _real_ default sort.

(That cache is also very annoying when you remove a column -_-' )

alexweissman:
Tablesorter also has an `updateCache` method: https://mottie.github.io/tablesorter/docs/example-update-all.html

PR #692, opened 2017-04-12
Fixed missing translator bug

splitt3r:
Found a stupid bug in one of my last commits :grin:


lcharette:
So it wasn't really broken? :P  
You should add the doc block above the translator while you're at it ;)

alexweissman:
I have been able to replicate this problem, so let's merge it in with the next hotfix.

PR #606, opened 2017-01-08
Added docker configuration

fullergalway:
Added an easy way to try out userfrosting by deploying with docker.

PR #696, opened 2017-04-19
Add docker-compose file for quick start.

colinmollenhour:
Inspired by #606 I set out to dockerize UserFrosting for a quick development environment. However, my implementation is quite a bit different from #606.


alexweissman:
Hey @colinmollenhour, do you have a lot of experience with Docker?  I'd like to discuss this in [chat](https://chat.userfrosting.com) to talk about deployment strategies more generally.  Could you join us when you get a chance?

PR #698, opened 2017-04-26
Replace default MySQL passwords, listen only on localhost

colinmollenhour:
This improves security by default for development environments by not exposing the ports to external interfaces and setting a more secure (but still hard-coded) database password.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,7 @@
+# See docker/README.md for usage instructions
 version: '2'
 services:
+
   php:
     build:
       context: ./docker/php
@@ -9,29 +11,31 @@ services:
       - DB_PORT=3306
       - DB_NAME=userfrosting
       - DB_USER=userfrosting
-      - DB_PASSWORD=password
+      - DB_PASSWORD=9rw46T1AucpTQrFs
     volumes:
       - .:/app
     links:
       - mysql
+
   nginx:
     build:
       context: ./docker/nginx
     ports:
-      - 8570:80
+      - 127.0.0.1:8570:80
     volumes:
       - .:/app
     links:
       - php
+
   mysql:
     image: mysql:5.7
     environment:
+      - MYSQL_ROOT_PASSWORD=JbIREUJlTI40Twft
       - MYSQL_DATABASE=userfrosting
-      - MYSQL_ROOT_PASSWORD=root
       - MYSQL_USER=userfrosting
-      - MYSQL_PASSWORD=password
+      - MYSQL_PASSWORD=9rw46T1AucpTQrFs
     ports:
-      - 8571:3306
+      - 127.0.0.1:8571:3306
     volumes:
       - userfrosting-db:/var/lib/mysql
 


alexweissman:
What's the point of having a different default password?

Silic0nS0ldier:
Changing the password won't offer any additional security, since its up on GitHub for all to see.

But since its containerised, the database can't be accessed from the outside. Plus to get in, they'd have to attain access to the server, at which point you've already lost.

colinmollenhour:
Yes, changing the listen address kinda circumvents the need to change password, I agree. One or the other is necessary, but not both so much.

However, even though the password is made publicly available, if a port scanner finds it and tries a quick dictionary attack, "root" would probably be one of the first 3 passwords it would try and then it would succeed. Port scanners also exist as viruses on local machines and local networks so I don't think it is bad to add a little extra security. Relative to "the internet" as a whole this docker-compose.yml file exists in an echo chamber so the password being publicly available doesn't completely mitigate it's usefulness. A serious production deployment should obviously change the default but in case it isn't so obvious or someone is just really lazy it is still a lot better than "root". :)

Silic0nS0ldier:
As mentioned in the [README.md](https://github.com/userfrosting/UserFrosting/tree/master/docker) file, the docker component is far from ready for production. In terms of actually making this secure, there is a lot that would need to be done. In its current state, changing configuration values on the repo won't do anything to improve security.

I will give though, that enforcing localhost only will help.

colinmollenhour:
Yes, I wrote the README file so I know what it says.. Just trying to make small improvements as I go. Clicking "Merge" takes the same amount of effort as clicking "Close"..

Silic0nS0ldier:
Huh. So you did. I've been off UF for a bit so I'm out of the loop. And for that exact reason I'll leave it up to @alexweissman .

alexweissman:
How about something besides `password`, but that is still clear that it's a value they should change?  Something like `changemebro`?

alexweissman:
@colinmollenhour I've added your Docker instructions to https://learn.userfrosting.com/installation/environment/docker.  Can you update the Docker config files (if necessary) and the docs for 4.1 when you have a chance?

colinmollenhour:
I'm not currently using/following the project so have no idea what if anything would need to be updated but I doubt if anything needs to be updated unless 4.1 has significant changes to the native installation instructions.

alexweissman:
A little bit - we moved `composer.json` one level up so that UF can be fetched with Composer, and we now have an all-in-one installation tool called [Bakery](https://learn.userfrosting.com/cli/commands).

I updated the `composer.json` path, but I'm not sure how the new installation command fits in with Docker since it handles `sprinkles.json`, installing npm and Bower deps, and running the migration.  The old piece-by-piece commands still work, so I am taking a conservative approach and leaving them the way it is for now.

Anyway, if you get back into UF, let us know!

Silic0nS0ldier:
This still needs to be addressed, though the time-frame is unclear (definitely closer than we were almost 2 years ago). One part of this can be integrated now however, listening to only localhost for inbound database connections.

PR #745, opened 2017-06-26
Fix for PostgreSQL setup

x00x70:
Stops charset errors at setup.

### DIFF
--- a/app/system/Bakery/Command/Setup.php
+++ b/app/system/Bakery/Command/Setup.php
@@ -112,7 +112,8 @@ class Setup extends BaseCommand
             'port' => $port,
             'database' => $name,
             'username' => $user,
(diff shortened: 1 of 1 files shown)


alexweissman:
`charset` should get pulled from the DB config rather than being hardcoded.

x00x70:
agree - just need bakery to read full config then ;)

alexweissman:
Alright, leaving this to you and @lcharette.  Taking this off the 4.1.3 roster.

lcharette:
Should be good to go. @x00x70 can you test to make sure it still works with Postgre? Should be fine with MySQL ;)

x00x70:
Tested and confirmed for PostgreSQL 9.6.2

PR #754, opened 2017-06-30
Fixed hardcode User class path.

freakinhuge:
Fixed hardcode User class path in the `AdminController` so that you can fully extend the `User` model.

### DIFF
--- a/app/sprinkles/admin/src/Controller/AdminController.php
+++ b/app/sprinkles/admin/src/Controller/AdminController.php
@@ -43,8 +43,11 @@ class AdminController extends SimpleController
             throw new ForbiddenException();
         }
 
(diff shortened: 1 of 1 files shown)


Silic0nS0ldier:
I suggest using [`$classMapper->staticMethod`](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Util/ClassMapper.php#L86) instead of `$classMapper->createInstance`. Otherwise there is an unnecessary increase in the overhead (as an empty `User` object ends up getting created).

alexweissman:
Thanks, this is now taken care of in 4.0.24 and 4.1.3 (https://github.com/userfrosting/UserFrosting/commit/aa43591ee0c26c7d928b32b2ec682de199c60e23).

Issue #763, opened 2017-07-09
Creating new User requires Email support


zharding:
When creating a user in the admin dashboard UF currently enforces the use of email. If your config is missing or incorrect the creation of the user fails. This can be an issue for development as the developer may have chosen to not setup email support. I do this frequently for a few reasons. 

**Solution**
Allow the user to be created with an option to not send an email. In which case a random password would be generated. A second possible solution, if email was opted out, create the user and leave the password blank. The user would remain inactive and the admin would have to manually (re)set the password and then enable the user. 

This would of course leave 'flag_verified' as negative and flag_enabled as negative to start. However, if email was opted out, simply mark 'flag_verified' as 1 for true. Manually enabling the account would set the 'flag_disabled' to 1. 

This would allow for for expansion into a 'customer' type model or simply easier on the developer lol.

alexweissman:
👍 

Please note that `flag_verified` has nothing to do with this, though.  Users created through the admin interface are _always_ automatically verified.

UserFrosting checks to make sure that a user does not have an empty password when they try to log in.  So, right now when an admin creates a new user, it sets the password to `''`, and the user gets an email which lets them set their password.

I think we could just have the option to manually set a password when the admin creates the account (actually, this used to be the _only_ way in V3 😱 )  We could bring this back, along with a "suggest password" tool that generates and displays a random value to the admin.  

Obviously this is not a good choice to use in production, but I can see how this would be useful in development/testing.

zharding:
https://github.com/userfrosting/UserFrosting/pull/764

lcharette:
As discussed in chat earlier, a solution would be to merge the existing "create user" with "change user password" form. When creating a user, the admin could select to enter a password or use the default email mechanism.

zharding:
I concur, 

This was something Alex and I discussed. I forgot to add the comment here. However, it has been incorporated into my PR here https://github.com/userfrosting/UserFrosting/pull/764

amosfolz:
Fixed in https://github.com/userfrosting/UserFrosting/pull/1017

PR #764, opened 2017-07-09
Admin - New user password option; set manually OR send email.

zharding:
The ability to create a user and manually set the password OR have a link sent to allow the user to set their own password. Also refactored the 'user-set-password' modal to separate the form and create a 'form/partials' This gives access to common form parts and allow the re-use of the 'user-set-password' controls.

**Original**
When creating a user manually there actually had to be a real email attached to the new user and the email settings must be configured and configured correctly. 

**Purpose**
Allow a developer to create a user without needing to send an email if so chosen. This also allows an admin to create a user with a non-existent email.

**Original Issue**
https://github.com/userfrosting/UserFrosting/issues/763


amosfolz:
Is there a reason this was never merged?

lcharette:
> Is there a reason this was never merged?

Note sure honestly...

amosfolz:
I think this can be closed as this functionality already exists. 
Looks like it was manually added in [this commit by Alex](https://github.com/userfrosting/UserFrosting/commit/ca72d35d7149f1f33096d0649fb7de3dfebd4cab#diff-8d8ccf778e9a0e551ad140119cf4dc04R1)

PR #765, opened 2017-07-10
Fixes error in password length validation error.

JoshNorris:


lcharette:
Note that has been fixed by @alexweissman in commit c9f60a8

alexweissman:
😁

PR #768, opened 2017-07-12
Missing permission check when uri_account_settings is not in role.

kevinrombach:
I have a role which is not allowed to access the account info page. But without this permission check the "My Account" link appears in the user-card.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,5 @@ public/css/min/*
 
 # Ignore personal config files (v0.3.x)
 app/config-userfrosting.php
+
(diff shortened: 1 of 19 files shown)


alexweissman:
Thanks!

alexweissman:
Ah crap, forgot to merge this in before the last release :-/

alexweissman:
Thanks, this has been merged in to https://github.com/userfrosting/UserFrosting/commit/1a320b21f692045d06fb98cf56443a54a8407a1a (v4.1.6).

PR #785, opened 2017-08-19
Fixed hard user deletion

splitt3r:
This https://github.com/laravel/framework/blob/5.4/src/Illuminate/Database/Eloquent/SoftDeletes.php#L33 method was causing an infinite loop on hard user deletion.

### DIFF
--- a/app/sprinkles/account/src/Database/Models/User.php
+++ b/app/sprinkles/account/src/Database/Models/User.php
@@ -161,7 +161,7 @@ class User extends Model
      * @param bool $hardDelete Set to true to completely remove the user and all associated objects.
      * @return bool true if the deletion was successful, false otherwise.
      */
-    public function delete($hardDelete = false)
+    public function deleteUser($hardDelete = false)
     {
         /** @var UserFrosting\Sprinkle\Core\Util\ClassMapper $classMapper */
         $classMapper = static::$ci->classMapper;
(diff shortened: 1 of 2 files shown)


app/sprinkles/account/src/Database/Models/User.php
```diff
@@ -180,10 +180,10 @@ public function delete($hardDelete = false)
             // TODO: remove any persistences
 
             // Delete the user
-            $result = parent::forceDelete();
```
splitt3r: https://github.com/laravel/framework/blob/4ab4e5f9470939e1c853196e1d5c84830c8b22e3/src/Illuminate/Database/Eloquent/Model.php#L1129

splitt3r: Otherwise the traits forceDelete Method calls the this method. See: https://github.com/laravel/framework/blob/5.4/src/Illuminate/Database/Eloquent/SoftDeletes.php#L33

alexweissman: Ok, got it.  It sounds like maybe overriding `User::delete` to support hard/soft deletes was a bad idea in the first place.  But do we need a `deleteUser` method either?  It sounds like maybe we should use [model events](https://laravel.com/docs/5.4/eloquent#events) to handle deleting/modifying related entities instead.  Choosing to do a hard/soft delete could be handled in the controller method.

PR #792, opened 2017-09-06
Several styling fixes

SarahBaghdadi:
- Chevron added to account settings dropdown menu to make it look menu-y
- Simplified logo in main-nav
- Table tool menu styles (gear--> bars)
- Separated "status" from "actions" in user table
- Hide hamburger menu icon for views where the side nav is already visible

These are all separate commits, so you should be able to cherry pick what you want.


app/sprinkles/admin/templates/tables/users.html.twig
```diff
@@ -72,25 +73,27 @@
...
+                        {% endverbatim %}{{translate("ACTIVE")}}{% verbatim %}
+                    </span>
                 {{/ifx }}
+            {{/ifx }}
+        </td>
+    </script>
+    <script id="user-table-column-actions" type="text/x-handlebars-template">
+        <td>
+            <div class="btn-group">
+                <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown">Edit <span class="caret"></span></button>
```
splitt3r: This https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/locale/en_US/messages.php#L99 should be used here.

SarahBaghdadi: Thanks.

lcharette:
I've review those changes:

- **Chevron added to account settings dropdown menu to make it look menu-y**
The base AdminLTE template doesn't have chevron, so that's why there wasn't any. Might be nice to have the chevron switch side when the menu is open, like the side menu.

- **Simplified logo in main-nav**
Removing the logo might be a regression or BC. I know I'll use one... But it does make more sense to not have one by default I guess.
capture d ecran 2017-09-17 a 19 37 09

- **Align nav logo left**
I prefer the site name centered (See screenshot above)

- **Hide hamburger menu link except when its needed**
I think we lost the ability to close the side menu on browser size windows?

- **Separated status from actions menu**
+1. Could probably further improve the button (make column same with as button, use "action" instead of "edit"? Or use an icon ? A simple gear (no button) could maybe more clean?). Also, bring back `action` as the column header?
capture d ecran 2017-09-17 a 19 44 12

- **Table tool menu styles**
+100. Way better.

- **Changed english translation of "Remember me!" to "Keep me signed in".**
+1

FYI, I'm not a big fan of using `id` for css styling. I might be wrong though.

app/sprinkles/core/assets/userfrosting/css/userfrosting.css
```diff
@@ -149,3 +161,45 @@ a .info-box {
...
+}
+
+@media(min-width: 768px) {
+    .main-header .logo {
+        text-align: left;
+    }
+}
+
+/* Padding for user menu */
+.user-menu a i {
```
alexweissman: why not just put it on `.user-menu .fa` or something?

SarahBaghdadi: ok

PR #800, opened 2017-10-10
Update messages.php

Rendername:
Changed "Панель" on "Панель управления", the correct RU full translation.


alexweissman:
Merged in v4.1.13-alpha.

PR #814, opened 2017-11-15
Refactoring on the Password class

enygma:
This updates the `Password` class to make a few changes:

- Moving the default cost value to a class constant and updating where it's used
- Replacing hash comparison with `hash_equals` to prevent timing attacks on equals comparison

### DIFF
--- a/app/sprinkles/account/src/Util/Password.php
+++ b/app/sprinkles/account/src/Util/Password.php
@@ -14,6 +14,8 @@ namespace UserFrosting\Sprinkle\Account\Util;
  */
 class Password
 {
+    const DEFAULT_COST = 12;
+
     /**
      * Returns the hashing type for a specified password hash.
      *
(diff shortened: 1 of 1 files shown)


app/sprinkles/account/src/Util/Password.php
```diff
@@ -26,7 +28,7 @@ public static function getHashType($password)
         // If the password in the db is 65 characters long, we have an sha1-hashed password.
         if (strlen($password) == 65) {
             return 'sha1';
-        } elseif (substr($password, 0, 7) == '$2y$12$') {
+        } elseif (substr($password, 0, 7) == '$2y$'.self::$_DEFAULT_COST.'$') {
```
alexweissman: Looks like there might be a typo here (should be `self::$DEFAULT_COST`)

alexweissman:
Looks pretty good!  Can you add a few tests (nothing fancy, just some basic I/O to test that inputted passwords are correctly matched to their hashes) in `sprinkles/account/tests/`?

alexweissman:
Ok, merged!  I've opened up #827 so we remember to rewrite this to make it more testable.

Issue #672, opened 2017-03-21
Config can't remove unwanted available locales


lcharette:
Setting the available locale in a sprinkle config file to remove the unwanted locale won't work. This only replace the locale displayed name.

```
            'locales' => [
                'available' => [
                    'fr_FR' => 'Français'
                ],
            ]
```

To solve this issue, available locale should only be an array of locale code (`'available' => ['fr_FR', 'en_US']`) and the locale name should be define elsewhere (locale metadata for example).

(Related to #671)

alexweissman:
As a short-term solution, what if `userfrosting/config` had a `getDefined` method that did the same thing as `get`, but then removed any keys from the sub-array that are assigned to `null`?

So then you could do:
```
            'locales' => [
                'available' => [
                    'fr_FR' => 'Français',
                    'en_US' => null,
                    'de_DE' => null,
                    ...
                ],
            ]
```

Then you could call `$config->getDefined('site.locales.available')` and it would strip out those null-valued keys.

lcharette:
I guess it could work. Any reason it couldn't be done in normal `get`? db/laravel settings?

alexweissman:
I mean, we could always do an `array_walk` or something to filter them out in the `config` service.

lcharette:
See #682 for a temp fix.

alexweissman:
The more general problem here is the ability for a schema/config/whatever file loaded in a later Sprinkle, to _remove_ (rather than simply overwrite) content from a previously loaded Sprinkle.

The question is, do we want to have something in the syntax of the files themselves that can do this?  Otherwise, I think my `getDefined` approach might be the best.

alexweissman:
I'm going to close this because it's _technically_ been fixed with the `Repository::getDefined` method.

The broader discussion about the correct way to store and load locale metadata can be continued in this issue: https://github.com/userfrosting/i18n/issues/4

Issue #637, opened 2017-02-04
Support for right to left (rtl) text/locales


lcharette:

alexweissman:
What was the issue again?  Just the formatting in views?

lcharette:
Something related to the end of sentence period and list bullet points not being on the right side for Arabic translation

PR #817, opened 2017-11-24
Hotfix development/added rtl support

aminakbari:
I added support for right-to-left languages for AdminLTE. All text will be written from right to left. Special CSS files will be imported from assest if the chosen language is right to left.


lcharette:
I'm not a big fan of https://github.com/userfrosting/UserFrosting/pull/817/files#diff-2f859228e39382fd31e49102c1e5162eR32, but I guess for now it will do. A more permanent solution would be to [use metadata](https://chat.userfrosting.com/channel/general?msg=CiJ2Y3ApHiMb4rGjE), but this needs to be compatible with #672. Also related to https://github.com/userfrosting/i18n/issues/4


This is a direct fix for #637

aminakbari:
Me neither. What we could be using for now is also to check if `translate("RTL")` is set or not.
Then, RTL languages only have to add a `"RTL" => "TRUE"` to their locale file.

lcharette:
That would indeed be sexier in the mean time. Can you open another issue for the metadata thing?

alexweissman:
There is one open in `i18n`: https://github.com/userfrosting/i18n/issues/4

alexweissman:
Gonna hold off on this until it gets cleaned up a little more.

PR #866, opened 2018-03-25
rowDelete event fix

Ekwav:
The `rowDelete` event from the `ufCollection` widget didn't include the deleted row.
But following the docs it should: https://learn.userfrosting.com/client-side-code/components/collections#deleterow


alexweissman:
Could someone do a clean rebase on this? It looks like it's trying to merge in all the commits from `develop` with it (which we don't want to do, if we want to release this as a hotfix for 4.1).

Ekwav:
I will just commit again to hotfix

Issue #776, opened 2017-07-28
Create User Without SMTP


pauljcurtis:
For development could a user be created without the requirement for SMTP settings. Could the 'email' go to logs instead?

alexweissman:
I think we could consider a general switch that shunts all mail to a log file instead, during development.

lcharette:
Would make the Vagrant install perfect... It's the only thing *not* set automatically in #829

lcharette:
Duplicate of #763.

PR #872, opened 2018-03-31
Add support for other config['mailer'] options.

apple314159:
PHPMailer supports other mailer options such as Mail, Qmail, and sendmail.
Tested with sendmail.

### DIFF
--- a/app/sprinkles/core/src/Mail/Mailer.php
+++ b/app/sprinkles/core/src/Mail/Mailer.php
@@ -44,15 +44,6 @@ class Mailer
 
         // Configuration options
         if (isset($config['mailer'])) {
-            if (!in_array($config['mailer'], ['smtp', 'mail', 'qmail', 'sendmail'])) {
-                throw new \phpmailerException("'mailer' must be one of 'smtp', 'mail', 'qmail', or 'sendmail'.");
-            }
---- a/app/sprinkles/core/src/Mail/Mailer.php
-+++ b/app/sprinkles/core/src/Mail/Mailer.php
-@@ -44,11 +44,17 @@ class Mailer
- 
-         // Configuration options
-         if (isset($config['mailer'])) {
             switch ($config['mailer']) {
             case 'mail':
                 $this->phpMailer->isMail();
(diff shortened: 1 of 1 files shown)


lcharette:
Related #776 ? Setting up `sendmail` as default, at least for Vagrant install, could be helpful. I'll have to check what email support Homestead as installed by default.

Either way, adding an `env` config for this would be nice too.

apple314159:
I don't use Vagrant or Homestead.

In our setup, our email server is on another machine. I wanted to queue email locally using sendmail to not block PHP.

app/sprinkles/core/src/Mail/Mailer.php
```diff
@@ -47,8 +47,23 @@ public function __construct($logger, $config = [])
             if (!in_array($config['mailer'], ['smtp', 'mail', 'qmail', 'sendmail'])) {
                 throw new \phpmailerException("'mailer' must be one of 'smtp', 'mail', 'qmail', or 'sendmail'.");
             }
-
-            if ($config['mailer'] == 'smtp') {
+--- a/app/sprinkles/core/src/Mail/Mailer.php
++++ b/app/sprinkles/core/src/Mail/Mailer.php
+@@ -44,11 +44,17 @@ class Mailer
```
lcharette: This shoudn't appear in the code.

app/sprinkles/core/src/Mail/Mailer.php
```diff
@@ -47,8 +47,23 @@ public function __construct($logger, $config = [])
...
+ 
+         // Configuration options
+         if (isset($config['mailer'])) {
```
lcharette: The if is not require since you have a `default` in the switch

app/sprinkles/core/src/Mail/Mailer.php
```diff
@@ -47,8 +47,23 @@ public function __construct($logger, $config = [])
             if (!in_array($config['mailer'], ['smtp', 'mail', 'qmail', 'sendmail'])) {
```
lcharette: This one is also not required anymore

lcharette:
@alexweissman That `if (isset($config['mailer'])) {`...  In the original code, if no ` mailer` was defined, the exception wasn't thrown. Was that intended? Otherwise, the switch will work fine if that ìf` is not there, but the ` hpmailerException` will always be thrown if there's no config thrown. 

Original : https://github.com/apple314159/UserFrosting/blob/master/app/sprinkles/core/src/Mail/Mailer.php#L46
Modified: https://github.com/apple314159/UserFrosting/blob/ccda4012456f0ea7aec6f6b110805c5736d0f622/app/sprinkles/core/src/Mail/Mailer.php#L46

lcharette:
Merged manually in f67e3582f217e4ced68b1b8c5ebab2691bf9f5c8 for the `develop` branch. Will be released in 4.2; Thanks !

PR #873, opened 2018-04-01
rowDelete event event fix

Ekwav:
The rowDelete event from the ufCollection widget didn't include the deleted row.
But following the docs it should: https://learn.userfrosting.com/client-side-code/components/collections#deleterow

### DIFF
--- a/app/sprinkles/core/assets/userfrosting/js/uf-collection.js
+++ b/app/sprinkles/core/assets/userfrosting/js/uf-collection.js
@@ -239,7 +239,7 @@
          */
          _deleteRow: function(row) {
              row.remove();
-             this.$element.trigger('rowDelete.ufCollection');
+             this.$element.trigger('rowDelete.ufCollection',row);
          },
          /**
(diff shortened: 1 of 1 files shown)


lcharette:
(That Travis error looks like a false negative...)

Ekwav:
It does work. Not sure if it should however. The other callbacks also just
take the `row`.

Alex Weissman <notifications@github.com> schrieb am So., 13. Mai 2018,
23:32:

alexweissman:
> Not sure if it should however. The other callbacks also just take the `row`

I don't understand...could you explain?

Ekwav:
The `row` variable is passed to the events `rowTouch.ufCollection` and `rowAdd.ufCollection` as well. https://github.com/Ekwav/UserFrosting/blob/0987216fb35feb8eb6a414a3a07e40542b1ec080/app/sprinkles/core/assets/userfrosting/js/uf-collection.js#L261

alexweissman:
Released in 4.1.18.

PR #875, opened 2018-04-09
Completed italian translations for every Sprinkle

optiroot:
Corrected every locale of every sprinkle to a correct and real italian translation.
As a native Italian speaker, I assure every string is correct.


lcharette:
Nice ! Thanks ! Will get merge in next update.

optiroot:
Thank you! :smile:

alexweissman:
Released in 4.1.18.

PR #884, opened 2018-05-28
Added cache control to fix bug with internet explorer not updating UF table data

aminakbari:
I was using UserFrosting to implement a software solution for a client of mine and it turned out that they were Internet Explorer fans. This was when I realized that UF tables did not work well in IE 11 and possibly Edge.
After debugging for some time, I came to the conclusion that this was because of the cache headers. I believe it is generally a good idea to set cache headers to disabled for API requests that would update frequently. So I set the cache headers for all Sprunje (Json) functions to "no-cache".
This fixed the IE problem which was a significant compatibility issue for UserFrosting.


Silic0nS0ldier:
`no-cache` isn't bullet proof BTW. Not sure about the specifics, but recently I hit some cases were caching didn't work. `no-store` would be a better choice as absolutely nothing is cached.

Silic0nS0ldier:
This has been superseeded by https://github.com/userfrosting/UserFrosting/pull/922

Issue #885, opened 2018-07-07
WhoopsRenderer not compatible with recent versions of Whoops error report template


alexweissman:
Our current `WhoopsRenderer` was written based on Whoops [2.1.8](https://github.com/filp/whoops/blob/2.1.8/src/Whoops/Resources/views/layout.html.php). Since then, newer versions of Whoops include a reference to a `$prettify` variable in `layout.html.php` that did not exist in Whoops 2.1.8.

Since we do not define this variable in `WhoopsRenderer`, this raises a notice:

> Undefined variable: prettify in D:\git\mingle\app\vendor\filp\whoops\src\Whoops\Resources\views\layout.html.php on line 28

I recommend that we either [define this variable in `WhoopsRenderer`](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php#L226-L229), or simply lock Whoops to version 2.1.8 in our `composer.json`.

silvioq:
Hi!

I patched my local code with this simple diff

```
diff --git a/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php b/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php
index 767ce1b..6edcd7e 100644
--- a/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php
+++ b/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php
@@ -227,6 +227,7 @@ class WhoopsRenderer extends ErrorRenderer
             "zepto"      => file_get_contents($zeptoFile),
             "clipboard"  => file_get_contents($clipboard),
             "javascript" => file_get_contents($jsFile),
+            'prettify' => true,
 
             // Template paths:
             "header"                     => $this->getResource("views/header.html.php"),
```

I don't know if this is the proper way for fix. If it's ok, I can make a PR.

alexweissman:
If this is working, then yes, please do!

lcharette:
I replicated this issue using the test. Should be good.

PR #886, opened 2018-07-22
Fix error template in WhoopsRenderer

silvioq:
- Resolve error template, adding prettify variable, which is required
  in template
- Add (very) simple test for WhoopsRenderer::render function
- Resolve userfrosting/UserFrosting#885


lcharette:
I reimplemented this fix in e8a6d8a in `hotfix`

silvioq:
Thanks, @lcharette

PR #904, opened 2018-10-13
Add route:list command

apple314159:


lcharette:
FYI, since the same thing/question was brought up with @ssnukala, if you add more commits to the same branch, it get added to the existing PR. no need to create a new one. 

I'll close this one for now ;)

PR #909, opened 2018-10-21
Updated ufForm.js to show Alerts on Success

ssnukala:
When the page is not reloaded after the ufForm submit, the page will show the alerts from the message stream. This is disabled by default (current functionality) can be enabled by updating the setting
value for showAlertOnSuccess to TRUE.


lcharette:
Committed directly into `develop` in 1c387ba3993d68683997fd0661aff422ba4e5062

PR #910, opened 2018-10-25
Avoid twig deprecation warning

silvioq:
Using "is defined" for prevent twig deprecation message:

`Silent display of undefined block "modal_footer" in template "modals/modal.html.twig" is deprecated since version 1.29 and will throw an exception in 2.0. Use the "block('modal_footer') is defined" expression to test for block existence`

https://github.com/twigphp/Twig/commit/ae9b503b17419102dbec3da4a583cf02c3ecf366

### DIFF
--- a/app/sprinkles/core/templates/modals/modal.html.twig
+++ b/app/sprinkles/core/templates/modals/modal.html.twig
@@ -1,8 +1,9 @@
 {# Base layout for modals.
 #}
 
-{# Conditional block.  See http://stackoverflow.com/a/13806784/2970321 #}
+{% if block('modal_footer') is defined %}
 {% set _modal_footer = block('modal_footer') %}
+{% endif %}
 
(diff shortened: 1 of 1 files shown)


lcharette:
Not sure why the `_modal_footer`  exist? Could we simply do : 

```
            {% if block('modal_footer') is defined %}
                <div class="modal-footer">
                    {{ block('modal_footer') | raw }}
                </div>
            {% endif %}
``` 

In any case, not why why the `default` is necessary. There's already an `if not empty` above it.

silvioq:
@lcharette , your solution is better, I will change my PR.

silvioq:
I created PR #911 ... you can close this Pull request.

lcharette:
> I created PR #911 ... you can close this Pull request.

FYI, you can add more commits to your fork to continue on the same PR, for futures one. Might be different if you post directly from GitHub...

PR #911, opened 2018-10-25
Avoid twig deprecation warning

silvioq:
Using "is defined" for prevent twig deprecation message:

`Silent display of undefined block "modal_footer" in template "modals/modal.html.twig" is deprecated since version 1.29 and will throw an exception in 2.0. Use the "block('modal_footer') is defined" expression to test for block existence`

https://github.com/twigphp/Twig/commit/ae9b503b17419102dbec3da4a583cf02c3ecf366


lcharette:
(Ref #910)

PR #922, opened 2019-01-12
Prevent caching of API routes with dynamic content

Silic0nS0ldier:
This is an improvement on https://github.com/userfrosting/UserFrosting/pull/884 to reduce the amount of reused code, and to make it easy to prevent caching of additional routes in a clear manner though the use of middleware.

There may be some routes that need to be excluded, and some added.

### DIFF
--- a/app/sprinkles/admin/src/Controller/GroupController.php
+++ b/app/sprinkles/admin/src/Controller/GroupController.php
@@ -289,8 +289,7 @@ class GroupController extends SimpleController
 
         // Be careful how you consume this data - it has not been escaped and contains untrusted user-supplied content.
         // For example, if you plan to insert it into an HTML DOM, you must escape it on the client side (or use client-side templating).
-		//set cache headers in order to stop specially IE to cache the result
-        return $sprunje->toResponse($response)->withHeader('Cache-Control', 'no-cache')->withHeader('Expires', '-1');
+        return $sprunje->toResponse($response);
     }
 
     /**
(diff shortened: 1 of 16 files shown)


app/sprinkles/admin/src/Controller/GroupController.php
```diff
@@ -524,8 +524,7 @@ public function getUsers(Request $request, Response $response, $args)
 
         // Be careful how you consume this data - it has not been escaped and contains untrusted user-supplied content.
         // For example, if you plan to insert it into an HTML DOM, you must escape it on the client side (or use client-side templating).
-        return $sprunje->toResponse($response);
-    }
+        return $sprunje->toResponse($response);    }
```
lcharette: Style on this line seams off

Silic0nS0ldier: Rather than play whack-a-mole with code styles I ran PHP CS Fixer. Normalised some things from previous commits.

Have to say, not a big fan of the PHPDoc rules. Namely the alignment. Makes internal documentation necessarily tedious, and the added spacing gets a little crazy when exception types are involved.

lcharette:
I haven't looked at all routes that this was applied to, but it looks legit. Middleware class may be better named or with a better namespace to indicate it's a Middleware instead of `UserFrosting\Sprinkle\Core\Util\NoCache`, but I don't think we have a specific "Middleware" location yet.

Silic0nS0ldier:
Yeah, I tried to match existing patterns, however the only example was `AuthGuard`. Definitely something to consider in the future though.

Issue #888, opened 2018-07-25
redirect failing in UserController::pageInfo when user not found


zbigcheese:
Recently it seems, named paths have stopped working.

Example route:

```
$app->group('/users', function () {
    $this->get('', 'UserFrosting\Sprinkle\Admin\Controller\UserController:pageList')
        ->setName('uri_users');
```

Example code from a controller:

```
$user = $this->getUserFromParams($args);
        // If the user no longer exists, forward to main user listing page
        if (!$user) {
            $usersPage = $this->ci->router->pathFor('uri_users');
            return $response->withRedirect($usersPage, 404);
        }
```

hitting the above point in a controller just produces a regular browser 404 page and does not redirect at all

alexweissman:
It would seem that our use of a 404 status code in a redirect is not compliant with (HTTP?) specs. See https://github.com/slimphp/Slim/issues/2355

I'm not sure why we did this originally, and I can't imagine why 404 would ever have been an acceptable status code for redirects.  At any rate, we should probably just change the code to 302.

lcharette:
Fixed in e7ac1d0 for 4.2

Tested and indeed 404 and 400 status code failed to redirect. Slim default to 302, so default it is now.

PR #929, opened 2019-01-25
Fix 404 for delete group from detail page

UrsBoller:
when deleting a group on route getInfo the group gets deleted. after success (200) app tries to reload the getInfo page where group doesn't exists anymore. therefore a redirect to main page is better than 404.

### DIFF
--- a/app/sprinkles/admin/src/Controller/GroupController.php
+++ b/app/sprinkles/admin/src/Controller/GroupController.php
@@ -228,7 +228,8 @@ class GroupController extends SimpleController
 
         // If the group doesn't exist, return 404
         if (!$group) {
-            throw new NotFoundException($request, $response);
+            $redirectPage = $this->ci->router->pathFor('uri_groups');
+            return $response->withRedirect($redirectPage, 301);
         }
 
         // Get group


Silic0nS0ldier:
I don't think having a redirect when there is no group found would be the correct way to handle this. We should be able to handle this before that initial refresh is even tried, and avoid the extra request altogether.

Silic0nS0ldier:
Agreed. Best part is it would maintain 100% backwards compatibility. Only thing I'm not sure about is how the returned data would be accessed. Does the UF Form plugin send response data to the event listeners?

Silic0nS0ldier:
It would appear that the redirect solution has already been implemented in 4.2.
#888 e7ac1d0

@lcharette ?

UrsBoller:
@lcharette @Silic0nS0ldier agree to close. it was a "short fix" to get rid of the 404 but indeed not the right place to fix. as you mentioned, the redirect should happen on clientside (onSuccess).
maybe there is a way to add a parameter to the modal which allows custom forwards on success.

UrsBoller:
@lcharette instead of "hardcoding" or "controller returns value", is there an option to expand the "form" when you call the modal? eg. when calling the modal to delete a group, add an 'onSuccess' value to the form to indicate where the user should be routed to, see [here](https://github.com/userfrosting/UserFrosting/blob/develop/app/sprinkles/admin/src/Controller/GroupController.php#L344). this would allow to call the same modal from different places and always tell the modal where the user should be forwarded to.

Silic0nS0ldier:
Could be imagining things, but I recall there already being a callback for this. If not, then it should be added. Lots of use cases for it.

lcharette:
@UrsBoller Adding the onSuccess url in the modal creation controller method does sounds like a good idea. 

Let's continue this discussion in #932.

PR #936, opened 2019-02-16
ClassMapper and Docker improvements

Silic0nS0ldier:
Original goal here was to improve readability of code when the class mapper is used. Unfortunately PHP 5.6's ability to handle compound statements is atrocious, so the scope of the changes has been severely limited and the incompatible changes have been pushed off to another 4.3 specific branch.

There area  few other tidbits in here as well (docker, README, etc) that should be non-functional changes for the most part.


Silic0nS0ldier:
Well, it seems that while PHP 5.6 supports the syntax, it doesn't support compound syntax. Unsupported changes will be split off to another branch in the meantime for <kbd>4.3</kbd>.

app/sprinkles/core/src/Util/ClassMapper.php
```diff
@@ -37,10 +37,7 @@ public function createInstance($identifier)
 
         $params = array_slice(func_get_args(), 1);
 
-        // We must use reflection in PHP < 5.6.  See http://stackoverflow.com/questions/8734522/dynamically-call-class-with-variable-number-of-parameters-in-the-constructor
-        $reflection = new \ReflectionClass($className);
-
-        return $reflection->newInstanceArgs($params);
+        return new $className(...$params);
```
lcharette: Wasn't that part that didn't work in PHP 5.6 ?

Silic0nS0ldier: Only when using it with compound statements (like a function result instead of a variable).

PR #937, opened 2019-02-16
PHP 7.1 Optimisations

Silic0nS0ldier:
There are numerous performance and usability improvements that simply were not possible under PHP 5.6. As we look ahead to UF 4.3 the hope is that this PR will serve as a spring board, so that when 4.2 stable is out we can hit the ground running.

Changes include;
- Increasing the minimum PHP version defined in `composer.json`.
- Removing PHP versions `5.6` and `7` from the Travis build matrix.
- Avoiding usage of `ClassMapper->staticMethod` in favour of working directly with the returned class name.


app/sprinkles/account/composer.json
```diff
@@ -13,8 +13,7 @@
     ],
     "require": {
         "birke/rememberme" : "^2.0",
-        "nikic/php-parser" : "^1",
-        "php": ">=5.6"
```
Silic0nS0ldier: what...

Silic0nS0ldier: Nevermind. Tricked myself. PHP requirement of internal sprinkles is determined by UF at the top level.

lcharette: This should probably removed, and only used in the master composer.json (unless a specific sprinkle requires a more higher PHP version)

PR #950, opened 2019-03-22
Correct welcome message in the italian locale

Poldovico:
I noticed the welcome message in Italian is using an incorrect variable for the user's name compared to the other locales, which resulted in the name not being displayed. This PR fixes that.


lcharette:
Thanks !

Issue #893, opened 2018-09-05
Download CSV filename error


SarahBaghdadi:
Hi all, the "Download CSV" feature generates a filename that doesn't open with my OS.  For example I downloaded the file:

`20180904-requests-sorts%5Bcreated_at%5D=desc&filters%5Bcompany_id%5D=2&properties%5B0%5D=company&properties%5B1%5D=courses&properties%5B2%5D=coursesNew&properties%5B3%5D=rate&properties%5B4%5D=paymentMethod&properties%5B5%5D=student.current.csv` 

generates the error:

`"Sorry, we couldn't find /Users/A/Downloads/20180904-requests-sorts%5Bcreated_at%5D=desc&filters%5Bcompany_id%5D=2&properties%5B0%5D=company&properties%5B1%5D=courses&properties%5B2%5D=coursesNew&properties%5B3%5D=rate&properties%5B4%5D=paymentMethod&properties%5B5%5D=student.cur. Is it possible it was moved, renamed or deleted?"`

The file itself is ok. If I change the filename to something else it opens and works fine. I'm using macOS 10.13.6 with Excel. 

I think the error is because of the length of the filename since when I reduce it by a few words it fixes the problem. If the filename was something short like `mytablename.csv` that would be a lot more legible.

alexweissman:
We need to truncate the generated filenames to 255 characters, or whatever the maximum limit is for common OSes.

amosfolz:
One of the difficulties with this is that maximum file name length does not just include the file name, but the length of the directory and any parent directories the file is stored in. For example, on Windows the default download folder is `C:\Users\Username\Downloads`. Username could be any length up to 20 characters (or more in some environments). 

 Without a way to know the length of the `username` or the default download path, I'm making the assumption to keep filename under 150 characters, leaving roughly 100 characters for the default storage path. 

I tried to keep it "pretty" by truncating to the nearest "&" before the 150th character, so the filename isn't just cut off at a random point. 

https://github.com/amosfolz/UserFrosting/commit/02b0fdf2917c67aa7cabf6c00efd01036aeb32e8

Let me know if this might work!

lcharette:
It's probably why most time you download a file, it's named `download` 🤕

amosfolz:
@alexweissman Can you provide any insight on why the file name was originally setup the way it was? I wonder if we can just make it simple and use `$filename = "$date-{$this->name}"`

amosfolz:
[...]

PR #954, opened 2019-04-03
Update Sprunje.php

amosfolz:
Fix for https://github.com/userfrosting/UserFrosting/issues/893
Should prevent file names from exceeding maximum allowed.
- Date is not really necessary as file metadata will have timestamp for creation. 
- Including the sorts/filters really made most file names overwhelming.

### DIFF
--- a/app/sprinkles/core/src/Sprunje/Sprunje.php
+++ b/app/sprinkles/core/src/Sprunje/Sprunje.php
@@ -188,8 +188,6 @@ abstract class Sprunje
             $result = $this->getCsv();
 
             // Prepare response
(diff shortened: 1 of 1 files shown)


app/sprinkles/core/src/Sprunje/Sprunje.php
```diff
@@ -190,7 +190,7 @@ public function toResponse(Response $response)
             // Prepare response
             $settings = http_build_query($this->options);
```
lcharette: So `$date` and `$settings` are not used anymore? We could removed it then.

Issue #870, opened 2018-03-29
Update to FontAwesome 5


lcharette:

haywirephoenix:
Download FA5, rename/delete and replace "userfrosting\app\sprinkles\core\assets\vendor\fontawesome" 

Change
`"vendor/fontawesome/css/font-awesome.css"`
`
"vendor/fontawesome/css/fontawesome.css",
                "vendor/fontawesome/css/all.css" `

in "userfrosting\app\sprinkles\core\asset-bundles.json"

done :)

lcharette:
Didn’t FA also changed icon names and prefix in 5?

lcharette:
> Download FA5, rename/delete and replace `"userfrosting\app\sprinkles\core\assets\vendor\fontawesome"`

Actually, for a permanent fix, it should be changed in the [bower](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/bower.json#L16) file.

We also need to provide support if someone wants to use the pro version (which I don't know how it works).

haywirephoenix:
The additional CSS files are supposed to add backwards compatibility.

Also note that they have removed the hyphen in font-awesome file and folder names

amosfolz:
I have been working on upgrading to FA 5 [here](https://github.com/amosfolz/UserFrosting/tree/fa5).

FA5 removed all aliases. As long as people are aware about the change it should not be a difficult transition. Probably easiest for people to just do the update and then go through their site looking for any missing icons, then update the FA tags as needed. 
[Here is a list](https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes) of all icons/aliases and any changes in FA5.

I'm not sure we could really make using FA 5 Pro any easier on the UserFrosting side. The user must decide between global/per-project setup and then add their authentication token.. See https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers

lcharette:
> It may be more appropriate for FA Pro to be added in a Sprinkle.

If we can't have a config for it, I guess so. 

Can we load with NPM both free and pro version, or does loading both will produce errors?

amosfolz:
> 
> 
> > It may be more appropriate for FA Pro to be added in a Sprinkle.
> 
> If we can't have a config for it, I guess so.
> 
> Can we load with NPM both free and pro version, or does loading both will produce errors?

You need to add special NPM registry entry with authentication key
 ```
[...]

PR #957, opened 2019-04-06
Update to Font Awesome 5 (Develop)

amosfolz:
Most of the changes are simply updating the 'fa' tag to the new 'fas' convention. 

I do not have a Font Awesome 5 Pro subscription so the bakery command still requires some testing. 

https://github.com/userfrosting/UserFrosting/issues/870


app/sprinkles/account/src/Database/Migrations/v400/GroupsTable.php
```diff
@@ -34,7 +34,7 @@ public function up()
                 $table->string('slug');
                 $table->string('name');
                 $table->text('description')->nullable();
-                $table->string('icon', 100)->nullable(false)->default('fa fa-user')->comment('The icon representing users in this group.');
+                $table->string('icon', 100)->nullable(false)->default('fas fa-user')->comment('The icon representing users in this group.');
```
Silic0nS0ldier: This will need to be done in a new migration.

app/sprinkles/core/src/Bakery/SetupFontAwesomeCommand.php
```diff
@@ -0,0 +1,68 @@
...
+
+        // Delete any Font Awesome settings in .npmrc but leave other lines untouched.
+        passthru("sed '/npm.fontawesome.com/d' -i .npmrc");
+
+        // Add the registry setting and new auth key to .npmc.
+        passthru("printf '%s\n' '@fortawesome:registry=https://npm.fontawesome.com/' '//npm.fontawesome.com/:_authToken=$newAuthKey' >>.npmrc");
+
+        // Install Font Awesome Pro Package.
+        $this->io->writeln('> <comment>npm install --save-dev @fortawesome/fontawesome-pro</comment>');
+        passthru('npm install --save-dev @fortawesome/fontawesome-pro');
```
Silic0nS0ldier: Not too sure about how this is handled. Will need to discuss this more elsewhere (and get some context regarding what the vision for the implementation is) before I can give any actionable feedback though.

app/sprinkles/account/src/Database/Migrations/v400/GroupsTable.php
amosfolz: The `fa` tag can still be used but documentation states it has been deprecated. Though I could not find anything definitive on when that tag will actually stop working.

Silic0nS0ldier: Will still need to update this separately (and I think update the fast-track migration hidden somewhere) for this to properly take.

lcharette: Yeah that might be problematic. A new migration for sure (otherwise existing setup won't have the change). At one point the default could be removed altogether from the db structure and be done in the controller.

lcharette:
Definitively something for `4.3.0` if it may break existing sprinkles.

app/sprinkles/account/src/Database/Migrations/v400/GroupsTable.php
amosfolz: > At one point the default could be removed altogether from the db structure and be done in the controller.

This would allow the use of other icon packages if for some reason the use of Font Awesome was not wanted at all.

lcharette: Yes indeed !

amosfolz: It looks like this is already being set in the controller here? 
https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/admin/src/Controller/GroupController.php#L386

lcharette: Indeed. While it would be better to set the default in the model, I don't see why a default value must be defined in the first place outside the controller.

app/sprinkles/account/src/Database/Migrations/v430/UpdateGroupsTable.php
```diff
@@ -0,0 +1,53 @@
...
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function up()
+    {
+        if ($this->schema->hasTable('groups')) {
+            $this->schema->table('groups', function (Blueprint $table) {
+                $table->string('icon', 100)->default('NULL')->change();
```
lcharette: Should it be null, or an empty string? 🤔

amosfolz: I went with NULL based off this post: https://stackoverflow.com/questions/38351498/remove-default-in-migration

I also compared to the other columns in the table that did not have a default set when the migration ran, and they all showed `NULL` for the `Default` property. In mysql here is how the results of my testing:
image

Silic0nS0ldier: Should be possible to do an export of the create table logic to see if the default value was truly reversed (the cleaner the better, leftovers attract edge cases).

Assuming I remember, I'll take a look at what the DDL ends up being.

amosfolz: Here is how default migration looks
userfrosting groups default

Here is the new migration `up`
UP

New migration `down`
DOWN

Silic0nS0ldier: I think the posted DDL is wrong. Or its mislabeled? I'll still have to look at this myself, but it works as a good reminder none-the-less, so thanks.

amosfolz: I am curious about what is wrong? 

Other than `('fas fa-user')` in [this line](https://github.com/amosfolz/UserFrosting/blob/fa5/app/sprinkles/account/src/Database/Migrations/v430/UpdateGroupsTable.php#L50), the down method returns the table to the 'original' state. I was not certain how to handle this but I thought it would be better to set the default to use the new `fas` convention rather than the `fa`.

I relabeled my pictures above that might help clear up any confusion.

amosfolz:
@Silic0nS0ldier What can I do to help make this review easier for you? 

Also, if it will help simplify things I can move the bakery command into a new branch and out of this PR.

Silic0nS0ldier:
@amosfolz My main limitation at the moment is time, or more specifically whenever I have free time there is something jumping up and down screaming for my attention. There isn't much that can be done about that in the meantime unfortunately, however I will get to this (eventually).

One thing that would help however is if you could note that a given PR is ready for review. If I see added commits without a message once the work is done, I'll just end up assuming that isn't in a ready for final review state.

amosfolz:
This is ready for review! :-)

app/sprinkles/account/src/Database/Migrations/v430/UpdateGroupsTable.php
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function down()
+    {
+        $this->schema->table('groups', function (Blueprint $table) {
+            $table->string('icon', 100)->default('fas fa-user')->nullable(false)->change();
```
lcharette: Shouldn't the default be `fa fa-user` here ?

app/sprinkles/core/src/Bakery/SetupFontAwesomeCommand.php
```diff
@@ -0,0 +1,69 @@
...
+
+        // Display nice explanation and ask which method to use
+        $this->io->write('This setup will configure Font Awesome 5 Pro. Font Awesome 5 Free is already included with UserFrosting.');
+
+        if (!$this->io->confirm('Would you like to continue?', false)) {
+            exit;
+        }
+
+        $this->io->note("The Authentication Key will be saved in `{$this->buildPath}/.npmrc`");
+        $this->io->write('Please enter your Font Awesome Authentication Key to begin. This will be used to download the neccessary packages.');
```
lcharette: This is actually called `Pro npm Package Token` in the Font Awesome account details.

amosfolz:
Looks like `5.9` was released in June. Might as well update to the newest version while this is a work in progress? https://fontawesome.com/changelog/latest

app/sprinkles/account/src/Database/Migrations/v430/UpdateGroupsTable.php
amosfolz: Fixed in c3dd512296e1b7d8566f8cddd8364da62620868c

app/sprinkles/core/src/Bakery/SetupFontAwesomeCommand.php
amosfolz: I removed the bakery command for simplicity sake. I noted the changes and have it saved in a different branch which I will work on sometime in the future...

amosfolz:
Not sure what happened with `AdminLTE.js` but I am going to blame the linter on my office computer for screwing up that file somehow... The error is fixed in b5d80605b8be7a66253a65bbb722ac376ac0b1a1.

Also needed to fix the following two icons:
`fa-gear` no longer exists, so using `fa-cog` as alternative.

`fa-trash-o` is now `fa-trash-alt`. Fixed these in 4ae024b3133fcaefd7a650250b50b6d9153b1f31

app/sprinkles/core/src/Bakery/SetupFontAwesomeCommand.php
amosfolz: This is more of a "nice to have" feature so I removed in 5a3d5467cf0cc4df1950554d806d79cb31f5f7f5 and will review it at a later date.

amosfolz:
> And pro doesn't work :(

Just leaving a note for myself if/when I come back to try to add in the bakery command...
I think the reason it didn't work is that you still need to manually update the `asset-bundle` with the pro file path and that was something that can't easily be done through bakery.

lcharette:
FA Pro is probably something we can live with having only a guide in the doc on how to do it in your sprinkle.

Issue #934, opened 2019-02-10
More details in CheckEnvironment.php


amosfolz:
If the folders `/log` `/cache` or `/session` have been accidentally deleted or moved, the error does not give helpful information. 

![clipboard - february 8 2019 1-55 pm](https://user-images.githubusercontent.com/33728190/52534957-4e6ff000-2d16-11e9-89bc-5c9cfab1571f.png)

Per Alex Weissman, "this ought to display the full path".

https://github.com/userfrosting/UserFrosting/blob/4dda14f611b6de5764e05d22538f5c179b0a18f7/app/sprinkles/core/src/Util/CheckEnvironment.php#L266-L274

lcharette:
Related #630.

amosfolz:
I understand this is part of some larger changes that are planned but I thought just changing the error message could help with this specific issue for now: 
https://github.com/amosfolz/UserFrosting/commit/6870f7a88cfdaad3cb266be49675f0a084c45d49

lcharette:
Note because of the `foreach` above, this has the potential of displaying the same message more than once. That's why the ` $file`  variable is used. The real fix would be to see why it contains `0` instead of the file name. I guess it has something to do with the locator returning an empty array.

amosfolz:
Ahh you bring up some good points. 

It seems the locator returns `0` if it can not find anything. Here is the result of renaming/deleting the `sessions` directory.
```
[2019-03-17 19:00:30] debug.DEBUG: shouldBeWriteable [] []
[2019-03-17 19:00:30] debug.DEBUG: Array
(
    [/home/vagrant/userfrosting/app/logs] => 1
    [/home/vagrant/userfrosting/app/cache] => 1
    [0] => 1
)
```

amosfolz:
Would it be appropriate to add a new function that checks for existence prior to running the `checkPermissions()`? It could be something like this:
```
        $shouldExist = [
        "log"     => $this->locator->findResource('log://'),
        "cache"   => $this->locator->findResource('cache://'),
        "session" => $this->locator->findResource('session://')
        ];

        foreach ($shouldExist as $directory => $path) {
          if ($path == null) {
            //Show error message for missing directory here. "$directory is missing..."
          }
        }
```

lcharette:
Those two different array are making thing complicated...

[...]

PR #958, opened 2019-04-06
More helpful error message in checkEnvironment.php

amosfolz:
Provide helpful error message in situation where `cache` `session` or `logs` is accidentally deleted or moved. 
The existing error was: 
image

This adds `checkDirectories` which provides a more helpful error message:
image

Because I was unable to come up with a more creative way to prevent `checkPermissions` from returning the generic error message, that check is skipped until the directories actually exist.

### DIFF
--- a/app/sprinkles/core/src/Util/CheckEnvironment.php
+++ b/app/sprinkles/core/src/Util/CheckEnvironment.php
@@ -125,6 +125,10 @@ class CheckEnvironment
             $problemsFound = true;
         }
 
+        if ($this->checkDirectories()) {
+            $problemsFound = true;
+        }
+
         if ($this->checkPermissions()) {
             $problemsFound = true;
         }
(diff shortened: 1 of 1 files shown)


app/sprinkles/core/src/Util/CheckEnvironment.php
```diff
@@ -125,6 +125,12 @@ public function checkAll()
             $problemsFound = true;
         }
 
+        if ($this->checkDirectories()) {
+            $problemsFound = true;
+        // Skip checkPermissions() if the required directories do not exist.
+        return $problemsFound;
```
lcharette: Careful with the indentation here

app/sprinkles/core/src/Util/CheckEnvironment.php
```diff
@@ -257,6 +263,39 @@ public function checkPdo()
...
+        $problemsFound = false;
+
+        $directoryPaths = [
+            "logs"     => $this->locator->findResource('log://'),
+            "cache"    => $this->locator->findResource('cache://'),
+            "sessions" => $this->locator->findResource('session://')
+        ];
+
+        foreach ($directoryPaths as $directory => $path) {
+          if ($path == null) {
```
lcharette: Indentation should be 4 spaces

app/sprinkles/core/src/Util/CheckEnvironment.php
```diff
@@ -257,6 +263,39 @@ public function checkPdo()
...
+            $problemsFound = true;
+            $this->resultsFailed['directory-' . $directory] = [
+                'title'   => "<i class='fa fa-file-o fa-fw'></i> A required directory was not found.",
+                'message' => "Please check that <code>userfrosting/app/$directory</code> exists.",
+                'success' => false
+            ];
+          }
+          else {
```
lcharette: Brace should be on the same line per PSR-2
https://www.php-fig.org/psr/psr-2/#51-if-elseif-else

lcharette:
Related #934

app/sprinkles/core/src/Util/CheckEnvironment.php
amosfolz: Thanks for the information! I will look over the PSR guides and get more familiar with the UF style guide as well.

lcharette:
Merged into hotfix for 4.2.1 release. Thanks !

Issue #867, opened 2018-03-28
Allow null group assignment for users


lcharette:
This will probably require a db migration to allow nullable value. The UI need to be allow for a "no group" option.

(Ref #843)

amosfolz:
Since this will require another DB migration I thought it might be good to include it in the same release as https://github.com/userfrosting/UserFrosting/pull/957 since that will require one as well. If there are no problems I will start working on it.

amosfolz:
I am running into some problems trying to figure out the best way to handle this. I am running into foreign key constraints from the relationships added here https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/account/src/Database/Migrations/v420/AddingForeignKeys.php#L44

Even if we make the `users.group_id` column nullable, we will not be able to save a value unless it also exists in the `groups.id` 
![image](https://user-images.githubusercontent.com/33728190/55841033-9c795800-5afb-11e9-9ee1-bf6d1a6b0b2d.png)

lcharette:
That's because the `group_id` is `0` while it should be `null`.

See : https://stackoverflow.com/a/15082906/445757

amosfolz:
> 
> 
> That's because the `group_id` is `0` while it should be `null`.
> 
> See : https://stackoverflow.com/a/15082906/445757

Can we use `value="0"` on the form, then in the controller do a check? Something like this right before the update is made:
 ```
        if ($data['group_id'] == 0) {
            $data['group_id'] = null;
        }
```
Otherwise validation fails if `group_id` is not a integer.

amosfolz:
Was addressed in https://github.com/userfrosting/UserFrosting/pull/964.

PR #964, opened 2019-04-11
Allow null group assignment for users (Issue #867)

amosfolz:
I think this is ready for a review. 
Maybe someone else can suggest a better way to handle showing an empty option using select2. 

Ref.: https://github.com/userfrosting/UserFrosting/issues/867


lcharette:
Looks like something is broken with the new migration : https://travis-ci.org/userfrosting/UserFrosting/jobs/518564703#L873

amosfolz:
I am going to try to see if I can figure out what is causing Travis to fail here. I noticed the migrations are all completing without any errors...
image

lcharette:
The good news is, you're not crazy. I checked out your branch and can't replicate the Travis issue. Even if I replicate Travis env locally (ie not using the `:memory:`database), the only issue I see is related to `testGetAsset`.

lcharette:
One thing to note, Travis complains about [`UpdateUsersTable.php:56`](https://travis-ci.org/userfrosting/UserFrosting/jobs/519717573#L967), which means the issue would be with the `down` method...

Actually, it could be a MySQL config thing. [The error is different on PgSQL](https://travis-ci.org/userfrosting/UserFrosting/jobs/519717583#L918). Something might be different between Travis MySQL setup and local/Homestead one.

amosfolz:
Okay, so I believe the PgSQL error is actually related to this https://github.com/userfrosting/assets/releases/tag/5.0.4

lcharette:
> Okay, so I believe the PgSQL error is actually related to this https://github.com/userfrosting/assets/releases/tag/5.0.4

And I believe it was fixed in `hotfix` here : https://github.com/userfrosting/UserFrosting/commit/36ed889c7ef197cf060f5a2d4f69e523aada94e8

Since this PR target UF 4.3, I suggest we leave it as is for now and look at it once we have released 4.2.1. We'll need to merge hotfix into develop at some point anyway.

amosfolz:
Is this something in the migration that needs to be fixed or is there possibly something in the test environment/configuration causing this?

amosfolz:
I think this will need to be a non-reversible change. I have been trying for hours to get a down migration to work that will set the column back to `not null` and have had no success.

lcharette:
> I think this will need to be a non-reversible change. I have been trying for hours to get a down migration to work that will set the column back to `not null` and have had no success.

If the change in the db is not reversed, and the files are reversed to 4.2.x, does it still works? Obviously, you won't be able to select "no group", but as long as the db still works, we should be good.

amosfolz:
> 
> 
> > I think this will need to be a non-reversible change. I have been trying for hours to get a down migration to work that will set the column back to `not null` and have had no success.
> 
> If the change in the db is not reversed, and the files are reversed to 4.2.x, does it still works? Obviously, you won't be able to select "no group", but as long as the db still works, we should be good.

It should still work.

If that happened the only difference would be `group_id` would remain `nullable`. 

```
   $table->unsignedInteger('group_id')->default(1)->comment('The id of the user group.')->change();
```

lcharette:
Only issue I see so far is in the user detail page : 

Capture d’écran, le 2019-07-10 à 20 52 24

The ` •` separating the username and group makes it weird now. Simple if in Twig should fix that.

app/sprinkles/admin/templates/forms/user.html.twig
```diff
@@ -24,7 +24,8 @@
                         {% if 'group' in form.fields.disabled %}
                             <input type="text" class="form-control" name="theme" value="{{user.group.name}}" disabled>
                         {% else %}
-                            <select id="input-group" class="form-control js-select2" name="group_id">
+                            <select id="input-group" class="form-control js-select2-group" name="group_id">
+                                <option value="0" </option>
```
lcharette: Is this valid html? Shouldn't it be `<option value="0"></option>` ?

app/sprinkles/admin/templates/forms/user.html.twig
```diff
@@ -24,7 +24,8 @@
                         {% if 'group' in form.fields.disabled %}
                             <input type="text" class="form-control" name="theme" value="{{user.group.name}}" disabled>
                         {% else %}
-                            <select id="input-group" class="form-control js-select2" name="group_id">
+                            <select id="input-group" class="form-control js-select2-group" name="group_id">
```
lcharette: `js-select2-group` is not defined in this PR, making the select not select2 styled. 

Plus, when applying select2, the "no group" styling is weird. Should probably says "no group" or something.

Capture d’écran, le 2019-07-10 à 21 26 15

amosfolz:
> 
> 
> Only issue I see so far is in the user detail page :
> 
> Capture d’écran, le 2019-07-10 à 20 52 24
> 
> The ` •` separating the username and group makes it weird now. Simple if in Twig should fix that.

Do we want this to show "No group" or just remove if they have that option select?

app/sprinkles/admin/templates/forms/user.html.twig
amosfolz: Fixed in b26b70512089f7f21cd3d09f609684edf95cf9c5

app/sprinkles/admin/templates/forms/user.html.twig
amosfolz: Fixed in b26b70512089f7f21cd3d09f609684edf95cf9c5

lcharette:
> Do we want this to show "No group" or just remove if they have that option select?

I think removing it by default makes more sense if someone doesn't want to use groups at all.

app/sprinkles/admin/templates/forms/user.html.twig
lcharette: The `No group`  will be the same formatting as the others...  Does Select2 support having a `<i>` inside the `option` element?

amosfolz:
> 
> 
> Only issue I see so far is in the user detail page :
> 
> Capture d’écran, le 2019-07-10 à 20 52 24
> 
> The ` •` separating the username and group makes it weird now. Simple if in Twig should fix that.

Fixed in 3467690d4e52b78b6c945d5dfc2c7b2f08a72677.


> 
> 
> > I think this will need to be a non-reversible change. I have been trying for hours to get a down migration to work that will set the column back to `not null` and have had no success.
> 
> If the change in the db is not reversed, and the files are reversed to 4.2.x, does it still works? Obviously, you won't be able to select "no group", but as long as the db still works, we should be good.

I retested and can confirm the DB still works fine in this situation.

lcharette:
> I retested and can confirm the DB still works fine in this situation.

I did too last and it does works fine.

app/sprinkles/admin/templates/forms/user.html.twig
amosfolz: Oops... It does not look like `<i>` is supported inside `option`

Issue #968, opened 2019-04-16
FeatureRequest: Hide locale select from UI if only one locale is available


avsdev-cw:
It would be handy if the locale select/display could be removed via a setting or if the system could detect when there is only 1 locale available and default to hiding the option to change it etc in that instance.

lcharette:
Kind of a duplicate of https://github.com/userfrosting/UserFrosting/issues/850

Unwanted locale can already be disabled in the config file. See https://learn.userfrosting.com/advanced/i18n#setting-up-the-site-language

I do like the idea of hiding the select if only one locale exist thought. But I'm not sure how complex it would be compare to the benefit of it considering the use of schema and default value...

avsdev-cw:
It kinda is, we already disable all locales bar en_US.

From our point of view its more of a "if the user doesn't need to use it, they don't need to know it exists" sorta thing.

I have started looking at it myself, but using overrides in another sprinkle. Once I started getting into the depths of it I quite quickly realised that actually there was some work needed in the UF default install. In some place the locale is hidden by a role, but others there is no wrap around it at all so there is a bit of inconsistency there already.

lcharette:
The admin UI controller has a `hidden` property. Same logic could be used in the normal user setting page controller. Locale field could then be added directly to this hidden array if the number of non null locale is less than 2.

avsdev-cw:
That's what I was thinking. I'll have to pull a new dev copy down and submit a PR if I get half a chance to look at it again.

lcharette:
Merged in hotfix for 4.2.1 release.

Issue #850, opened 2018-02-07
Metadata for locale 


lcharette:
See https://github.com/userfrosting/i18n/issues/4

This is dumb, as it doesn't works if the `default` has a space somewhere : https://github.com/userfrosting/UserFrosting/blob/develop/app/sprinkles/core/src/ServicesProvider/ServicesProvider.php#L437

This is rubbish and that's not how the default works : https://github.com/userfrosting/UserFrosting/blob/develop/app/sprinkles/core/src/ServicesProvider/ServicesProvider.php#L451

And that whole thing is just messy : https://github.com/userfrosting/UserFrosting/blob/develop/app/sprinkles/core/config/default.php#L140-L159

**tl;dr** FU the current `available` config, FU the `default` config with it's obscure order. Locale metadata with dependencies and name FTW. An array of available locale FTW.

lcharette:
So first question. How do we store locales config and where ?

In a Yaml file? In `/locale/fr_FR/fr_FR.yaml` or in `locales/fr_FR.yaml` ? In core sprinkle? A new sprinkle add a new locale / wants to replace the config they replace said file in their own `locales/` folder ?

lcharette:
This as been implemented with new version of i18n for UF 4.4

PR #969, opened 2019-04-18
Hide locale field when only 1 locale is available (#968)

avsdev-cw:


avsdev-cw:
FYI - no conflicts caused by merging with hotfix branch either (currently on develop branch)

app/sprinkles/account/src/Controller/AccountController.php
```diff
@@ -658,8 +668,18 @@ public function pageSettings(Request $request, Response $response, $args)
         // Get a list of all locales
         $locales = $config->getDefined('site.locales.available');
 
+        // Hide the locale field if there is only 1 locale available
+        $fields = [
+            'hidden'   => [],
+            'disabled' => []
+        ];
+        if (count($config->getDefined('site.locales.available')) <= 1) {
```
lcharette: Could reuse `$locales` here. Could be mirrored on line 519 for consistency.

lcharette:
As this doesn't introduce any breaking changes, I think it's safe to merge into hotfix.

app/sprinkles/account/src/Controller/AccountController.php
avsdev-cw: Would it be more preferable to use: 
```php
        $localePathBuilder = $this->ci->localePathBuilder;

        // Get locale information
        $currentLocales = $localePathBuilder->getLocales();
```
vs
```php
        // Get a list of all locales
        $locales = $config->getDefined('site.locales.available');
```

in all cases?

lcharette: I'm not 100% sure, but I think `localePathBuilder` will only return the locale associated with the one currently in use? It will at least return them in the inheritance order.

lcharette: In any case, this behavior will be changed in #850, so I guess it doesn't matters for now.

lcharette:
So apparently I forgot to change the branch before merging... I'll see what I can do 🤔

lcharette:
Ok, fixed. Merged in hotfix for 4.2.1 release. Thanks !

Issue #832, opened 2017-12-17
Blacklist common passwords


alexweissman:
Together with minimum length, [blacklisting common passwords](https://github.com/danielmiessler/SecLists/tree/master/Passwords) is the most effective known password policy.

Not sure how many we should blacklist by default.

Probably, we should also enforce an entropy requirement and blacklist `password = username` and `password = email`.  See https://blog.codinghorror.com/password-rules-are-bullshit/

Silic0nS0ldier:
Probably (definitely?) overkill, but https://haveibeenpwned.com/Passwords has a known exposed passwords list. All encoded in SHA1 (cause there are people out there that have passwords that identify them 🤦‍♂️). Pity it has no frequency with them.

splitt3r:
@Silic0nS0ldier awesome idea. https://github.com/nextcloud/password_policy/pull/60 This is a good examplte / an easy way of getting started.

Silic0nS0ldier:
Since you've reminded me, there is a password manager that has since integrated with the service. And based on Troy Hunt's blog posts, the API endpoints are highly optimised (most requests don't get past the CDN which Cloudflare footed the bill of) so any potential bandwidth theft implications are essentially eliminated.

Hmm... What would be the best place to perform the check? Server-side or client-side?

lcharette:
If you rely on a 3rd party API, we should consider caching the info on our side using the cache service. Then it's either a question or doing the check in either form validation (client or server side).

Silic0nS0ldier:
Update regarding the API: only anonymous requests will be accepted in the future.

The anonymous API works by giving the first few characters of the hash, and receiving hashes that match in return (minus the characters that were sent originally).

Silic0nS0ldier:
On the topic of caching, we'd need some system to keep the size to a reasonable amount. We could also limit caching to common results, since each hashed password is paired with a number indicating the number of items its appeared. I'm sure one of Troy's blogs has a graph we can use to get an idea of what a good number would be.

Silic0nS0ldier:
Some things to consider are mentioned here https://twitter.com/alexjamesbrown/status/1063552695133487104

amosfolz:
> 
> 
> Probably (definitely?) overkill, but https://haveibeenpwned.com/Passwords has a known exposed passwords list. All encoded in SHA1 (cause there are people out there that have passwords that identify them 🤦‍♂️). Pity it has no frequency with them.

[...]

PR #974, opened 2019-05-06
Have I Been Pwned Integration (Password security enhancements) #832

amosfolz:
This will need more work but I feel is at a place to open up for some feedback. 
The proposed feature adds two additional `config` values under `site.password_security`.
```
'password_security' => [
             'enforce_no_compromised'   => '5', // Set to false to turn off this feature. Otherwise, provide a numeric string, which sets the maximum number
                                       // of times that is "acceptable" for a password to have appeared in breaches. The recommended and most secure
                                       // option is '0' - meaning only passwords that are not on the list of compromised passwords will be allowed.

             'enforce_update_passwords' => true // Settings this to true will check user's passwords against list of known compromised passwords at each log on.
]
```
(I am open to suggestions on the naming convention for these).


lcharette (commented):
The controller is really long... While having private function in the controller is not wrong, I think it would make more sense to put both `checkPassword` and `checkHash` in a separate (static) class? Would make easier to use those methods elsewhere as well as writing proper tests.

amosfolz:
> 
> 
> The controller is really long... While having private function in the controller is not wrong, I think it would make more sense to put both `checkPassword` and `checkHash` in a separate (static) class? Would make easier to use those methods elsewhere as well as writing proper tests.

I thought they would probably need to be moved. I will need some guidance on where to move them to...

Silic0nS0ldier:
Looks like there are some changes on master that aren't on develop. Probably want to sync it up. Would make the diff clearer.

app/sprinkles/account/config/default.php
```diff
@@ -113,7 +113,14 @@
...
                 ]
             ]
-        ]
+        ],
+        'password_security' => [
+             'enforce_no_compromised'   => '5', // Set to false to turn off this feature. Otherwise, provide a numeric string, which sets the maximum number
+                                       // of times that is "acceptable" for a password to have appeared in breaches. The recommended and most secure
+                                       // option is '0' - meaning only passwords that are not on the list of compromised passwords will be allowed.
+
+             'enforce_update_passwords' => true // Settings this to true will check user's passwords against list of known compromised passwords at each log on.
```
Silic0nS0ldier: I think this would better belong under the login group above. Comment is important, however semantically this isn't where you would expect a login specific setting to reside. At least that is my take on it.

app/sprinkles/account/locale/en_US/messages.php
```diff
@@ -134,6 +134,9 @@
             'PAGE'              => 'Choose a new password for your account.',
             'SEND'              => 'Set New Password and Sign In'
         ],
+        'SECURITY'  => [
+            'RESET_REQUIRED'    => 'The password security policy does not allow use of known compromised passwords. Either this feature has been activated after you originally set up your accounty or your password has recently been added to the list.'
```
Silic0nS0ldier: Key name is a little too generic. It is foreseeable that we could add new criteria for mandating a password reset in the future (such as changes to required password complexity and sitewide password resets). Don't have any suggestions right now, but if anything comes to me I'll let you know.

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Generates a list of compromised passwords stored in cache or by querying Have I Been Pwned API.
+     *
+     * First check the cache to see if the hash prefix is stored.
+     * If not found in cache, query Have I Been Pwned API and store response in cache.
+     * @see https://haveibeenpwned.com/API/v2
+     * @param  string $password
+     * @return array  $result An array containing the password checked and the number of breaches.
```
Silic0nS0ldier: Return doesn't need or care about the variable that is returned, just the type and any documentation. Some documentation around the data shape would also be appreciated. Having been dealing with deeply nested associated arrays for the past few months, it would be awesome if we could come up with a way to avoid it if possible.

Really wish phpDoc supported the concept of a tuple here. But alas, what could have been. https://github.com/phpDocumentor/fig-standards/pull/115/files

Potential value of using something like a class however is that the shape is inherently documented and there is much greater potential for later expansion. Probably a good idea to take a step back here and consider the best way to handle this, as I feel the helper classes could ultimately be integrated into this theoretical result class. Would be a good idea to have more discussion here I think.

app/sprinkles/account/src/Controller/AccountController.php
```diff
@@ -37,6 +38,28 @@
...
+     * AuthGuard: false
+     * Route: /account/check-password
+     * Route Name: {none}
+     * Request Type: POST
+     * @see https://haveibeenpwned.com/API/v2
+     * @param Request  $request
+     * @param Response $response
+     * @param array    $args
+     */
+    public function checkPasswordApi(Request $request, Response $response, $args)
```
Silic0nS0ldier: Not too sure about the name of this function. I think `checkPassword` would suffice, and revising the comment to be open to later expansion is worth considering. In a similar vein to comment above re the result type and if a class should be considered, the response shape may also need some revision.

I haven't checked, but its also worth checking that if UF has a standard API response shape that it is being reflected here (I thinking the answer is no currently).

app/sprinkles/account/src/Controller/AccountController.php
```diff
@@ -405,6 +428,36 @@ public function login(Request $request, Response $response, $args)
 
         $currentUser = $authenticator->attempt(($isEmail ? 'email' : 'user_name'), $userIdentifier, $data['password'], $data['rememberme']);
 
+        // Check if the enforced password update setting is configured.
```
Silic0nS0ldier: As @lcharette has noted, this is a bit lengthy. Would be worth factoring this out to some common method. Same could possible be said for regular password resets, but I'll need to check the source in those areas first to determine if there is a gain to be made.

Silic0nS0ldier (changes requested):
Few things to be addressed here. There will likely need to be some discussion in most areas first though.

app/sprinkles/account/src/Controller/AccountController.php
```diff
@@ -500,52 +428,42 @@ public function login(Request $request, Response $response, $args)
...
-            exit;
-        } else {
-            $ms->addMessageTranslated('success', 'WELCOME', $currentUser->export());
-
-            // Set redirect, if relevant
-            $redirectOnLogin = $this->ci->get('redirect.onLogin');
-
-            return $redirectOnLogin($request, $response, $args);
-        }
+        return $redirectOnLogin($request, $response, $args);
```
amosfolz: Any ideas on why this would `return` twice? @lcharette said I need an exorcist. 
This line is still getting returned after line 457 above is sent. (Sometimes it works correctly when I test it) See screenshot below. 

image

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
amosfolz: > Having been dealing with deeply nested associated arrays for the past few months, it would be awesome if we could come up with a way to avoid it if possible.

Now that you mention it, there is not really any need to be returning an array. We can simply return a string with the number of breaches. The calling method already has the password being checked so there is no reason to be passing it back and forth.

app/sprinkles/account/config/default.php
amosfolz: I think you are right and logically `enforce_update_passwords` would be in the login group. However, `enforce_no_compromised` is not necessarily a login specific setting... 

Any suggestions on better key names for these two items?

app/sprinkles/account/src/Controller/AccountController.php
amosfolz: > Not too sure about the name of this function. I think checkPassword would suffice,

I am not sure, is it bad practice to have two different functions in different classes with the same name? 

> and revising the comment to be open to later expansion is worth considering.

How about something like: 
``` 
Checks password use eligibility. 
```

app/sprinkles/account/config/default.php
Silic0nS0ldier: I'm thinking something along the lines of having the `password_security` group define criteria passwords must adhere to (e.g. complexity customisation, exposures, etc) for the sake of extend-ability in the future. So basically the rest stays as is.

app/sprinkles/account/locale/en_US/messages.php
Silic0nS0ldier: Maybe something like `SECURITY.RESET_REQUIRED.COMPROMISED`? Deeper nesting would allow for additional security related reset requirements such as `SECURITY.RESET_REQUIRED.INSUFFICENT_COMPLEXITY`, and `SECURITY.RESET_REQUIRED.EXPIRED` (for mandating required routine password changes) or whatever the individual using the platform wants.

Makes it easier to drop features in without having breaking changes too.

app/sprinkles/account/config/default.php
amosfolz: I changed `enforce_update_passwords` to `enforce_reset_compromised` and moved it under the login group. Although `update` would work, `reset` is more in alignment with the existing verbiage.

app/sprinkles/account/locale/en_US/messages.php
amosfolz: Great idea. I added this change in the recent commits.

app/sprinkles/account/src/Controller/AccountController.php
amosfolz: I updated the name to `checkPassword` and believe I have addressed the other ideas as well.

app/sprinkles/account/src/Controller/AccountController.php
amosfolz: I added methods to `PasswordSecurity` to make these checks. I think it cleaned up the code a lot but if my implementation is not what you had in mind please let me know.

app/sprinkles/account/src/Controller/AccountController.php
Silic0nS0ldier: Not sure I follow. It should be impossible for a `return` statement to be triggered twice unless the client makes 2 requests. Some reproduction steps plus expected and actual outcomes might help.

app/sprinkles/account/assets/userfrosting/js/uf-password-security.js
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+
+  jQuery.validator.addMethod("check-password", function(value, element, breaches) {
+    return breaches < breachThreshold;
+  }, jQuery.validator.format("The password security policy does not allow use of known compromised passwords. Please enter a different password."));
+
+  if (breachThreshold != -1) {
+    //user is "finished typing," do something
+    function doneTyping() {
```
Silic0nS0ldier: I can't see any reason for this to be wrapped in a conditional. Currently I can see this cause JS errors under certain scenarios as well.

app/sprinkles/account/config/default.php
```diff
@@ -99,7 +99,8 @@
     */
     'site' => [
         'login' => [
-            'enable_email' => true // Set to false to allow login by username only
+            'enable_email'              => true, // Set to false to allow login by username only
+            'enforce_reset_compromised' => true // Settings this to true will check user's passwords against list of known compromised passwords at each log on. If this option is enabled then enforce_no_compromised should also be enabled.
```
Silic0nS0ldier: Suggested alternate comment
```
// Setting this to true will force users to reset their password if it considered compromised at login.
```

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
```diff
@@ -0,0 +1,138 @@
...
+        $this->config = $config;
+    }
+
+    /**
+     * Generates a list of compromised passwords stored in cache or by querying Have I Been Pwned API.
+     *
+     * First check the cache to see if the hash prefix is stored.
+     * If not found in cache, query Have I Been Pwned API and store response in cache.
+     * @param  string $password
+     * @return string A numeric string representing the number of times a password has been compromised.
```
Silic0nS0ldier: A string seems odd here. Representing how?

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
```diff
@@ -0,0 +1,138 @@
...
+            return $this->getHashArrayFromAPI($hashPrefix);
+        });
+
+        return $this->checkHash($passwordHash, $hashArray);
+    }
+
+    /**
+     * Checks the maximum number of times that is acceptable for a password to have appeared in breaches.
+     *
+     * @return string Numeric string with -1 meaning disabled.
```
Silic0nS0ldier: Shouldn't this technically be an integer? Would be worth casting it as such too.

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        return $this->config['site.login.enforce_reset_compromised'];
+    }
+
+    /**
+     * Check a password SHA1 hash against an array of compromised password hashes.
+     *
+     * @param  string $hash  The hash of the potential password to be used.
+     * @param  array  $array Array of password hashes in the format c2d18a7d49b0d4260769eb03d027066d29a:181 - or <hash>:<number of breaches.
+     * @return string A numeric string representing the number of times a password has been compromised.
```
Silic0nS0ldier: As earlier, why is this a string?

Silic0nS0ldier:
I've added some comments (but neglected to use the code review function, good job me). Mainly some concerns around return types and potential for script errors.

app/sprinkles/account/assets/userfrosting/js/uf-password-security.js
amosfolz: If this feature is disabled by setting `site.password_security.enforce_no_compromised.breaches` we do not want to run the client-side validation to run.

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
amosfolz: There is no specific reason other than being a string by 'default'. I will update it to an integer.

app/sprinkles/account/src/Controller/AccountController.php
```diff
@@ -405,6 +429,38 @@ public function login(Request $request, Response $response, $args)
...
+                'page' => [
+                  'validators' => [
+                      'set_password'    => $validator->rules('json', false)
+                  ]
+                ]
+              ], $token);
+
+                $ms->addMessageTranslated('info', 'PASSWORD.SECURITY.RESET_REQUIRED.COMPROMISED');
+
+                return $response->withRedirect($forcePasswordChange);
```
amosfolz: @Silic0nS0ldier I'm going to expand on this here since the code changed a lot since my first comment. 

**Expected behavior:** when `site.login.enforce_reset_compromised` is set to `true` then the user's password is checked at each log in. If the password is found to be unacceptable according to the config values, the user is expected to be redirected to the password reset page `return $response->withRedirect($forcePasswordChange);` and the `login` function should exit.

**Actual behavior:** The controller is returning both of the return statements at L460 and L469. You can see this in the Fiddler screenshot and based off that fact that the message returned here at L460 ends up displaying on the homepage in the screen shot below. 
```
return $response->withRedirect($forcePasswordChange);

        return $redirectOnLogin($request, $response, $args);
```


image

image


To reproduce, you should be able to checkout this branch and try to log in a user with a "bad" password, such as `password1234`

Silic0nS0ldier: Thanks. I've managed to reproduce the issue so I'll into this further once I have the chance. Depending on how tricky this proves to be, I should have an answer before the end of my work break.

Silic0nS0ldier: Looks to be some issues in client side scripts causing this.

From what I can see, what should be happening is that when someone logs in with an exposed credential they are redirected to the set password page? If yes then that raises the question of what is the best way to handle this. If we have evidence of exposed credentials would be not want to require the user to verify their identity via email and outright reject login?

amosfolz: > From what I can see, what should be happening is that when someone logs in with an exposed credential they are redirected to the set password page?

Yes this is how I intended for it to work. 

> If yes then that raises the question of what is the best way to handle this. If we have evidence of exposed credentials would be not want to require the user to verify their identity via email and outright reject login?

Well, maybe. There are at least two situations to consider.
1. This feature is activated on an existing UserFrosting site and as a result a large number of users may be required to update their passwords. 

2. The feature is already active on a site and a new breach is added to Have I Been Pwned list, flagging a user's password as now being unacceptable for use. 

In either situation it does not necessarily mean that the UserFrosting site's username/passwords have been exposed, just that in some other breach the password has been exposed.

I do understand your point about verifying their identity though. Forcing the user to verify their identity would be the more secure option, if only slightly more secure. As always it comes down to whether user experience or security is more important.

 If we approach this as a standard password rule, would it be normal to verify a user before allowing them to update their password if say, the minimum password length was raised from 8 characters to 10?

app/sprinkles/account/assets/userfrosting/js/uf-password-security.js
amosfolz: My JS skills are not great. If there is another way to handle this I am happy to change it. Otherwise, I do not know of any other way to handle this.

app/sprinkles/account/src/Controller/AccountController.php
Silic0nS0ldier: I think a situation where there is a possibility that someones password has been exposed we would want to explicitly **prevent** a quick and easy reset. Otherwise we would just be helping a would be attacker to lock a user out of their account. The danger goes up a lot of the account hijacked is that of the admin.

Other scenarios we could implement later like password expiration or changes to complexity requirements would be part of a different discussion.

amosfolz: What you are suggesting will diminish user experience and I am not convinced it will actually result in better security. Right now, users can reset their passwords via the account settings page. If there is benefit from forcing a user to verify their identity prior to resetting their password we should not be allowing password resets in the application at all - only via password reset email. 

Consider that there is a data breach of some company and an email and password are exposed by that breach that are also being used in a UF application. By the time said data breach actually makes its way to being listed on HIBP site (After being announced by breached company or attackers, and then undergoing HIBP verification, vetting, etc..) the attacker has already had time to log into the UF application and reset the user's password via the User Settings page.


> Other scenarios we could implement later like password expiration or changes to complexity requirements would be part of a different discussion.

I am not trying to discuss implementing those. I consider the password blacklist to fall into the same category as those password constraints/requirements, thus asking the question: 
> If we approach this as a standard password rule, would it be normal to verify a user before allowing them to update their password if say, the minimum password length was raised from 8 characters to 10?

I think we could use some outside input on this. I am happy to head in the proposed direction if there is consensus it is the best option. 
@lcharette @alexweissman

amosfolz: According to [this article](https://www.owasp.org/index.php/Testing_for_weak_password_change_or_reset_functionalities_(OTG-AUTHN-009)), and IF this feature falls under the category of a `password change` rather than a `password reset`, the question to ask is:

> - Is the old password requested to complete the change?

> The most insecure scenario here is if the application permits the change of the password without requesting the current password. Indeed if an attacker is able to take control of a valid session they could easily change the victim's password.

Though 'm not entirely sure whether redirecting the user to the page after they login covers the requirements of "requiring current password to allow password change".

lcharette (changes requested):
As discussed in #993, The config should be `site => password => security`

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
amosfolz: Changed to `int` with new comment

amosfolz:
> 
> 
> As discussed in #993, The config should be `site => password => security`

This is updated.

amosfolz:
This is ready for review. 

I have changed the process. If a password is found on the list of compromised passwords `flag_password_reset_required` is set for that user. This flag can easily be set using additional methods in the future and should allow for customization. 

A check was added to the login process. If a user has `flag_password_reset_required` they will be denied login until their password is reset. (They are directed to a similar page as `forgot-password` so they can enter their email and begin the password reset)

lcharette:
> For example, if we wanted to add capability for admin to force a password reset for a user.

༼ つ ◕_◕ ༽つ

Silic0nS0ldier:
The experience around the (root) admin account being on a compromised passwords list could use some work. Namely, the admin is not forced to perform a password reset to login despite the recorded user activity stating that they have to. Probably not great that the account with the most access isn't covered by these additional protections.

amosfolz:
> 
> 
> The experience around the (root) admin account being on a compromised passwords list could use some work. Namely, the admin is not forced to perform a password reset to login despite the recorded user activity stating that they have to. Probably not great that the account with the most access isn't covered by these additional protections.

Good catch! This is because the `if` that checks if `flag_password_reset_required` is set was using an instance of `currentUser` from prior to that field being set. (so it wasn't enforcing the change until the next login.) This should be fixed now.

amosfolz:
I believe this is ready for another review.

app/sprinkles/account/assets/userfrosting/js/uf-password-security.js
```diff
@@ -0,0 +1,45 @@
+$(document).ready(function() {
```
Silic0nS0ldier: There is a potential error in here (a function will be called despite not being defined if no compromised passwords is disabled). Also, try to keep string quotes consistent. There isn't any linting for this currently, but we should strive for consistency.

Logic here should be something like this.

```js
$(document).ready(function() {

    var breachThreshold = site.password.security.enforce_no_compromised.breaches;

    // Only perform security checks if they are enabled
    if (breachThreshold != -1) {
        // Setup before functions
        var passwordInput = $('form"'.find('input[name=password]');
        var typingTimer; // Timer identifier
        var doneTypingInterval = 200; // Time in ms

        /**
         * Checks if password is valid according to site policies.
         */
        function triggerPasswordValidation() {
            var ajaxData = {};
            ajaxData['password'] = $(passwordInput).val();
            ajaxData[site.csrf.keys.name] = site.csrf.name;
            ajaxData[site.csrf.keys.value] = site.csrf.value;
            $.ajax({
                type: 'POST',
                url: site.uri.public + '/account/check-password',
                data: ajaxData,
                success: function(data) {

                    $(passwordInput).rules('add', {
                        'check-password': data
                    });

                    $(passwordInput).valid();
                }
            });
        }

        // On keyup, start the countdown
        passwordInput.keyup(function() {
            clearTimeout(typingTimer);
            if (passwordInput.val()) {
                typingTimer = setTimeout(triggerPasswordValidation, doneTypingInterval);
            }
        });


        jQuery.validator.addMethod(
            'check-password',
            function(value, element, breaches) {
                return breaches <= breachThreshold;
            },
            jQuery.validator.format('The password security policy does not allow use of passwords that have been exposed by a prior data breach. Please enter a different password.')
        );

    }
});
```

Silic0nS0ldier (changes requested):
A few issues, but the core logic looks good. I haven't tested it this time however.

app/sprinkles/account/src/Database/Migrations/v430/AddPasswordRestColumnUsersTable.php
```diff
@@ -0,0 +1,57 @@
...
+ * See https://laravel.com/docs/5.4/migrations#tables
+ *
+ * @author Amos Folz
+ */
+class AddPasswordRestColumnUsersTable extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public static $dependencies = [
```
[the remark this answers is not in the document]
amosfolz: I think we could. If this is the preference I should probably update the other two "modifying" migrations in `4.3`.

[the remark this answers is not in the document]
amosfolz: @lcharette Should I change this and the other migrations that were already merged?

app/sprinkles/account/src/Database/Migrations/v430/AddPasswordRestColumnUsersTable.php
```diff
@@ -0,0 +1,57 @@
...
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function up()
+    {
+        if ($this->schema->hasTable('users')) {
+            $this->schema->table('users', function (Blueprint $table) {
+                $table->boolean('flag_password_reset_required')->default(0)->comment('Set to 1 to force user to reset their password, 0 otherwise.')->after('flag_enabled');
```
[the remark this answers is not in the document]
amosfolz: This formatting is easier to read. But as far as I can tell the common formatting throughout migrations is to use a single line for each column in the table. 

This style format is used in some controllers for DB queries and such though.

app/sprinkles/account/src/Database/Migrations/v430/AddPasswordRestColumnUsersTable.php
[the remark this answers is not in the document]
lcharette: Yes. Both will work, but reflection is easier to work with in IDE

app/sprinkles/account/assets/userfrosting/js/uf-password-security.js
amosfolz: lol... I'm going to quote you when this was inside a conditional, as you're suggesting it be again: 

> I can't see any reason for this to be wrapped in a conditional. Currently I can see this cause JS errors under certain scenarios as well.

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * Queries Have I been Pwned API to generate list of hashed compromised password suffixes.
+     *
+     * @param string $hashPrefix The prefix (first 5 characters) of hashed password.
+     *
+     * @return array Array of password hashes in the format c2d18a7d49b0d4260769eb03d027066d29a:181 - or <hash>:<number of breaches.
+     */
+    private function getHashArrayFromAPI($hashPrefix)
```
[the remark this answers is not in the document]
amosfolz: You're right. 

I'll think about the best way to handle this. We probably want to try to query HIBP at least a few times before allowing user to login and bypass the check...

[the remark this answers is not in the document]
Silic0nS0ldier: Got to be careful with retry logic. At scale, it can translate to a massive spike in requests for an API.
Possible scenarios include;
* API returns 500 error due to a fault, could be an edge case or everything, retry logic here would easily result in a larger number of requests (CloudFlare wouldn't cache these either, so it would fairly consistently hit the function itself until potentially DDoS protection gets triggered).
* Response codes indicate that retries should not be attempted exist, we would be wasting our own time retrying these.
* Connection timeouts could be attributed to failures anywhere along the connection (local misconfiguration, API hit its own response timeout and terminated).
* Deliberate rate limiting by HIBP (if a UF site were to become heavily used, we could end up hitting this)

Ultimately, it may come down to not retrying at all and logging that the check failed. HIBP passwords should be supplemental to the security such that any UF site doesn't get busted by any issues.

app/sprinkles/account/src/Database/Migrations/v430/AddPasswordRestColumnUsersTable.php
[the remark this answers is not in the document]
Silic0nS0ldier: Probably worth discussing, implemented in PHP CS (or whatever the thing we use is called), and documented in the style guide.

This can be addressed outside of this PR however, so I'm fine with this for now. #1020

app/sprinkles/account/src/Database/Migrations/v430/AddPasswordRestColumnUsersTable.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Users table migration
+ * Adds a `flag_password_reset_required` column to the users table.
+ * Version 4.3.0.
+ *
+ * See https://laravel.com/docs/5.4/migrations#tables
+ *
+ * @author Amos Folz
+ */
+class AddPasswordRestColumnUsersTable extends Migration
```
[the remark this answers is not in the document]
amosfolz: Fixed in e985aee4d80605c1684bb2708385dfea15763378

app/sprinkles/account/assets/userfrosting/js/uf-password-security.js
amosfolz: Fixed in e985aee4d80605c1684bb2708385dfea15763378

app/sprinkles/account/src/Database/Migrations/v430/AddPasswordRestColumnUsersTable.php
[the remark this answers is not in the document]
amosfolz: It seems a leading `\` is being added somewhere inside the `migrator`. 
```
 [ERROR] Unfulfillable migrations found ::
         => \UserFrosting\Sprinkle\Account\Database\Migrations\v430\AddPasswordResetColumnUsersTable (Missing dependency
         : UserFrosting\Sprinkle\Account\Database\Migrations\v400\GroupsTable)
```
Some changes will likely need to be made in the migrator if we want to make this work. Unless we want to try to make those changes for `4.3`, should probably just hold off on this for now.

app/sprinkles/account/src/Authenticate/PasswordSecurity.php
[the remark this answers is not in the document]
amosfolz: I've tried to fix this in the recent commits. (Also switched over to Guzzle)

app/sprinkles/account/src/Repository/PasswordResetRepository.php
```diff
@@ -33,7 +33,9 @@ class PasswordResetRepository extends TokenRepository
     protected function updateUser(UserInterface $user, $args)
     {
         $user->password = Password::hash($args['password']);
+        $user->flag_password_reset_required = false;
         // TODO: generate user activity? or do this in controller?
```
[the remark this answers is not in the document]
amosfolz: It doesn't look like it is happening in controller or here, so I would say no.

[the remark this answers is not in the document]
amosfolz: It would make sense to log some type of message "User set a new password".

Then the controller would be responsible for logging activity messages related to why a user needs to reset their password.

lcharette:
FYI, I tagged this to 4.4.x as I think it will require more testing from both @Silic0nS0ldier and I.

Silic0nS0ldier:
As can be seen, I'm trying to resurrect this PR nearly 2 years on. Its been synced with master and had some issues resolved (mainly PHP crashes relating to recently completed deprecations and type violations).

lcharette:
Closing this, as it need to be updated for UF5. It should be implement as it's own sprinkle anyway.

PR #976, opened 2019-05-07
Respects CSRF_ENABLED environment variable

Poldovico:
The previous statement would evaluate to true for any value of CSRF_ENABLED.
Using the strict comparison operator means if we set the variable to any false-evaluating values other then boolean false (0, '0', '0.0', '' and so on), then CSRF will be disabled.
getenv() evaluates to boolean false if the environment variable is not set, so I know of no simple way to distinguish between the variable being unset, in which case we want to default to enabling CSRF, and it being explicitly set to boolean false.

### DIFF
--- a/app/sprinkles/core/config/default.php
+++ b/app/sprinkles/core/config/default.php
@@ -96,7 +96,7 @@ return [
         * Note : CSRF Middleware should only be disabled for dev or debug purposes.
         */
         'csrf' => [
-            'enabled'          => getenv('CSRF_ENABLED') ?: true,
+            'enabled'          => (getenv('CSRF_ENABLED') !== false) ? getenv('CSRF_ENABLED') : true,
             'name'             => 'csrf',
             'storage_limit'    => 200,
             'strength'         => 16,


app/sprinkles/core/config/default.php
```diff
@@ -96,7 +96,7 @@
         * Note : CSRF Middleware should only be disabled for dev or debug purposes.
         */
         'csrf' => [
-            'enabled'          => getenv('CSRF_ENABLED') ?: true,
+            'enabled'          => (getenv('CSRF_ENABLED') !== false) ? getenv('CSRF_ENABLED') : true,
```
Silic0nS0ldier: Wouldn't something like `getenv('CSRF_ENABLED') !== false` suffice here? Feels like that would clamp down on more edge cases as well.

Poldovico: unfortunately getenv() returns false for undefined environment variables, so this would make CSRF disabled by default unless the environment enables it specifically.
I strongly dislike my own solution, but I can't come up with anything less inelegant.

Poldovico: The ideal scenario might be changing the variable itself to be CSRF_DISABLED, so if we don't set it it's enabled.

lcharette: > The ideal scenario might be changing the variable itself to be CSRF_DISABLED, so if we don't set it it's enabled.

That would be a breaking change that would required to wait for version 4.3.0. 
I think it's safe enough to keep it this way for now

alexweissman: The solution here is to not use the native `getenv` at all, but rather Laravel's [`env` helper](https://laravel.com/docs/7.x/helpers#method-env) which takes a default value as its second parameter. So this can just be `env('CSRF_ENABLED', true)`. This also has the nice side effect of allowing falsy values like `false` in the environment variable.

PR #980, opened 2019-05-15
Hide .env file and allow caching of JSON under NGINX

zhangchenglin:
QQ截图20190515180521

/userfrosting/public/assets-raw/userfrosting/favicons/manifest.json

### DIFF
--- a/webserver-configs/nginx.conf
+++ b/webserver-configs/nginx.conf
@@ -80,6 +80,12 @@ server {
     }
     ## End - Let's Encrypt
 
+    ## Begin - Prevent file content downloads from leaking
+    location ~ .*\.(env)?$ {
+        deny all;
+    }
+    ## End - Prevent file content downloads from leaking
+
     ## Begin - Handle PHP requests
     location ~ \.(php)$ {
         # Throw away any requests to execute PHP scripts in other directories
--- a/webserver-configs/nginx.conf
+++ b/webserver-configs/nginx.conf
@@ -105,7 +105,7 @@ server {
     ## End - Handle PHP requests
 
     ## Begin - Caching static files
-    location ~* \.(png|gif|jpg|jpeg|svg|ico|css|js|woff|ttf|otf|woff2|eot)$ {
+    location ~* \.(png|gif|jpg|jpeg|svg|ico|css|js|woff|ttf|otf|woff2|eot|json)$ {
         include /etc/nginx/mime.types;
         expires max;
 


lcharette:
FYI, style Failure can be ignored

(This one will need review from someone more familiar than me about nginx config...)

Silic0nS0ldier:
@zhangchenglin Good contributions from what I can see, but could you update the description to explain what this PR is aiming to achieve? Much easier to perform an effective code review if we know what the problem/intended change is. Can prove to be very useful if the source is ever the subject of additional changes later as well.

Silic0nS0ldier:
@zhangchenglin While I can understand the change to support caching of JSON, the configuration should not need to care about the `.env` file. If properly configured, the only files accessible by should be those inside the `public` folder (minus `index.php` which is handled specially).

If you are having this issue, there are 2 issues that need to be checked.
1. Is your server vulnerable to directory traversal attacks.
2. Are you storing `.env` files anywhere that would result in their exposure (e.g. `public` or `app/sprinkles/*/assets`

webserver-configs/nginx.conf
```diff
@@ -80,6 +80,12 @@ server {
     }
     ## End - Let's Encrypt
 
+    ## Begin - Prevent file content downloads from leaking
```
Silic0nS0ldier: As mentioned in my last comment, this should not be necessary on a properly configured server.

lcharette:
> If you are having this issue, there are 2 issues that need to be checked.

Or your host is pointing to `/` instead of `public/`...

webserver-configs/nginx.conf
zhangchenglin: > As mentioned in my last comment, this should not be necessary on a properly configured server.


The correct configuration you said is the default configuration? 
I recognize that you are talking about the correct configuration, this is not necessary, but the information in this file is indeed critical and important, so I think a few lines of code can be changed to annotation form. You can be prompted to use UserFrosting, this file needs to be done in case of some incorrect configuration.

avsdev-cw: Can you provide an example URL where you are able to get to this file (with reference to localhost)?  If there is a way to get to it with default configuration it is a serious security issue that would need to be addressed.

On a side note, UserFrosting is not meant to be installed under /var/www/public_html (or similar) which is the default root of nginx/apache etc. Rather, it should be installed somewhere like /usr/local/userfrosting, /usr/share/nginx/userfrosting (as used as the example in the nginx.conf file) or /var/www/userfrosting and then a configuration for nginx should be made with a root/directory alias at /usr/local/userfrosting/public, /usr/share/nginx/userfrosting/public, or /var/www/userfrosting/public using the 3 previous example locations.

zhangchenglin: > Can you provide an example URL where you are able to get to this file (with reference to localhost)? If there is a way to get to it with default configuration it may be a serious security issue that would need to be addressed.
> 
> On a side note, UserFrosting is not meant to be installed under /var/www/public_html (or similar) which is the default root of nginx/apache etc. Rather, it should be installed somewhere like /usr/local/userfrosting, /usr/share/nginx/userfrosting (as used as the example in the nginx.conf file) or /var/www/userfrosting and then a configuration for nginx should be made with a root/directory alias at /usr/local/userfrosting/public, /usr/share/nginx/userfrosting/public, or /var/www/userfrosting/public using the 3 previous example locations.

Sorry, JSON files should not be cached.

Silic0nS0ldier:
As per previous discussion, only the `public` folder should be exposed to the internet. Because of this file restrictions are not necessary since the expectation is that nothing sensitive be stored in this directory (technically including `index.php`, which is only hidden citing it being an implementation detail that would otherwise creating undesirable URLs).

Closing.

Issue #868, opened 2018-03-28
Locale selector for translator


lcharette:
The translator service can only work with one defined locale. We need a way to force a locale, for example when sending an email to a non-user. We can't simply use the logged-in user locale in that case.

lcharette:
I started looking into this and I need your feedback on something. Here's my current plan. Basically, it's one locale per translator instance : 

```
$translator = new Translator('EN_us');
$translator->translate('HOME'); // Home

$translator_fr = new Translator('FR_fr');
$translator_fr->translate('HOME'); // Accueil
```

In this case, the Translator object can load and cache all keys for the selected locale. It will load all files for the defined locale, and would be able to return the locale metadata (#850; eg. `$translator->getLocaleAuthor()`) and so on. But it does change the whole behavior of the current translator, which the constructor accept [an array of keys](https://github.com/userfrosting/i18n/blob/c964f2dd4bb4ca66a1ea61ea94bf51cf0844a2d9/src/MessageTranslator.php#L40) generated by the [LocalePathBuilder](https://github.com/userfrosting/i18n/blob/c964f2dd4bb4ca66a1ea61ea94bf51cf0844a2d9/src/LocalePathBuilder.php#L35).

This solution would effectively merge the `Translator` and the `LocalePathBuilder` into the same class, but the Translator service would then only be able to translate one locale. So translating into a different locale would require a new instance (like the example above), or use a method to switch the overall locale (eg. `$translator->loadLocale('ES_es');`)

amosfolz:
This would mean any missing or incomplete keys would no longer fall back to a "base" locale?

lcharette:
No, as the fallback locale will be defined in the metadata.

amosfolz:
I think it's a good plan then. Would the client-side code still work the same?

lcharette:
Goal is for `->translate(...)` to stay the same

lcharette:
This as been implemented with new version of i18n for UF 4.4

Issue #671, opened 2017-03-20
Tool for creating/updating language files


alexweissman:
I think it would make it a lot easier for translators, if we had some kind of interactive tool for creating language files, as well as updating them when changes are made to `en_US`.

We also have quite a few languages in UF3 that still need to be ported over to UF4, and I think this would be easier to get done if we had a tool:

- Spanish
- Dutch
- Italian
- Turkish
- Romanian

amosfolz:
Is the idea for the bakery to be used as the _interactive tool?_.

lcharette:
I think that was Alex initial idea. 

I think the first thing we need is a way to generate a list of keys available in English, but not in X language. #850 & #868 might be need first.

lcharette:
A new Bakery command, `locale:compare`, has been added for UF 4.4.0. This will makes it easier to detect values that needs to be translated. Not a fully interactive tool, but I think it's enough for now. See other comment here about future commands we could add : https://github.com/userfrosting/UserFrosting/pull/992#issuecomment-578563058

PR #992, opened 2019-06-04
Locale:* bakery command

amosfolz:
Adds three Bakery commands:
- `locale:missing-keys` check for missing locale `keys`
- `locale:missing-values` check for empty/duplicate locale `values`.
- `locale:fix-keys` Fix locale keys by adding any missing key:values from another locale. I.e. add all key:values that exist in `en_US` locale files (but not in `es_ES` files) into `es_ES` locale files. 

`locale:fix-keys` can also be used to create the folder structure and all locale files for a brand new locale.


lcharette:
Related #671

amosfolz:
> 
> 
> Also, quick note about that code vs testing. You have a `foreach` in a `foreach` in an `if` in a `foreach`... This will probably make it hard to create a meaningful test for that class. You should look at separating the code into more protected methods to avoid this nesting. One example, the `table` io generation should be in it's own method. This would also means it could probably be reused between both commands.
> 
> (And yes, I know. [`MigrateStatusCommand`](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Bakery/MigrateStatusCommand.php) is the opposite of what I just said :P )

I will need some help on how to do this in a different way. Basically this entire command revolves around a whole lot of arrays...

amosfolz:
Below is a picture of what the code currently produces: 
image

Or, you can just compare two locales:

image

app/sprinkles/core/src/Bakery/LocaleFixKeysCommand.php
```diff
@@ -0,0 +1,205 @@
...
+     */
+    protected function fix($base, $alt, $filePath)
+    {
+        //If the directory does not exist we need to create it recursively.
+        if (!file_exists(dirname($filePath))) {
+            mkdir(dirname($filePath), 0777, true);
+        }
+
+        // Build the respository and then merge in each locale file.
+<<<<<<< HEAD
```
lcharette: You have some merging conflict here...

amosfolz:
I am not sure why Travis failed as this PR has nothing to do with `userControllerTest`...

lcharette:
False negative. I relaunched the test

amosfolz:
This is ready for another review.

lcharette:
I've made some changes in 8e13cc7. Most changes are style related or adding the PHP 7 type hint for method arguments and return values. I've also optimized the code to make use of return value where possible instead of class properties. Also FYI, class properties where marked as **static** (eg. `protected static $path;`), which is not necessary in this case (they were not called statically). 

In the bigger changes:
- There was a edge case in `missing-keys` where a key containing an array wouldn't display as missing if the target locale had the key, but containing a string. 
- I changed the `locale:fix-keys` options from `-f` to `-l`. `f` should be associated with `force` whenever possible. In this case, I added the force option so the confirmation can be skipped.

Let me know if some of my changes doesn't give the right result, or if you don't understand some of them (might be my personal coding preference). 

Before merging this in, I'll try to add some tests for those three commands.

lcharette:
I've been thinking about this command since my last commit and I think the whole logic of comparing the files is flawed. 

The commands you created compares the content of `locale/en_US/messages.php` with `locale/es_ES/messages.php` for example. The problem is comparing files doesn't represent the end result which merge all arrays from all files. It is great to do single file comparison, but let's say the Spanish translation is in `locale/es_ES/mensajes.php`. The commands will show all keys from the English as missing, but the translation will actually works in UF.

I could also create false positive if the core translation has a key that contains an array (`'FOO' => array(...);`), but a sprinkle overwrite that (`'FOO' => 'BAR';`).

I think it would be best if the translator would have a feature to return a list of all registered keys. Since 4.4 Milestone focus on translation improvement, I'll push this feature to 4.4. At the same time, improvements to the `i18n` package (#868) will probably help with this, and it also means the comparison can probably be moved to i18n and the translator itself.

amosfolz:
I think this entire feature has had a lot of creep outside what I was really trying to accomplish. As per @alexweissman original post in #671 , this was primarily meant to be a tool for helping to keep the built-in translation files up-to-date. 

> The commands you created compares the content of locale/en_US/messages.php with locale/es_ES/messages.php for example. The problem is comparing files doesn't represent the end result which merge all arrays from all files. It is great to do single file comparison, but let's say the Spanish translation is in locale/es_ES/mensajes.php. The commands will show all keys from the English as missing, but the translation will actually works in UF.

This is true but I'm not sure it's necessarily a flaw. I think when you write code there are assumptions that have to be made. In this situation the assumption is that a standardized, logical naming schema will be used for translation files. It seems it would get extremely complicated to write code that will update/create the translation files without these simple file naming conventions in place. 

> I think it would be best if the translator would have a feature to return a list of all registered keys. Since 4.4 Milestone focus on translation improvement, I'll push this feature to 4.4. At the same time, improvements to the i18n package (#868) will probably help with this, and it also means the comparison can probably be moved to i18n and the translator itself.

I'm fine with this but I still think https://github.com/userfrosting/UserFrosting/pull/1006 needs to be considered for `4.3`. This will allow translators to update any missing translations and can be added in `4.3.*`

lcharette:
Maybe both method of comparing a locale could live? 🤔 

> I'm fine with this but I still think #1006 needs to be considered for `4.3`. This will allow translators to update any missing translations and can be added in `4.3.*`

I agree. Changed #1006 to 4.3 milestone.

amosfolz:
> I think it would be best if the translator would have a feature to return a list of all registered keys. 

I think this could pretty easily be done by adding a method or two to the `MessageTranslator` class.
For example:
```
        $locator = $this->ci->locator;
        $builder = new LocalePathBuilder($locator, 'locale://', [en_US]);
        $loader = new ArrayFileLoader($builder->buildPaths());
        $translator = new MessageTranslator($loader->load());
```
$translator will have an array of all `en_US` keys.

We could just add a method to return [`$translator->items`](http://api.userfrosting.com/source-class-UserFrosting.I18n.MessageTranslator.html#42), which would be an array of all `en_US` keys/values in this example. 

I actually looked back at my notes and this was my original approach for comparison between two different locales, but ended up changing it because because I couldn't see a way to find what files had what keys defined.

lcharette:
That would probably work great with #868, so you can get a list of keys directly : 

```
$keys = $translator->getKeys();
$FrKeys = $trsanslator->getKeys('fr_FR');
```

lcharette:
@amosfolz I've reworked your command into a single `locale:compare` command. Having `missing-keys` and `missing-values` was weird. The compare command list all difference between the two locales. Obviously, the new command makes uses of the new methods available in `i18n` to list all keys (and the actual comparison has been moved to `i18n` also).

Regarding the `locale:fix` command, I don't think it's a good fit yet. The points I didn't like was the use of `php-cs-fixer` inside the code (makes it required in the non-dev composer packages) and the fact the file are actually modified (which required the necessary permissions).

What I propose, is we add in future version an `output` argument to the `compare:locale`, and eventually create a new command, which will output the difference, or the whole locale, in a specific version (display, cvs, txt, array, etc.)

PR #993, opened 2019-06-05
Config settings for password min/max length

avsdev-cw:
Created some config settings for setting the min/max lengths of the password.

Due to some limitations in UserFrosting/Fortify there is no way to add the config setting name into the YAML and have it update via substitution. Where a loaded YAML is used with min/max password lengths there is an additional 2 lines per password field which updates the loaded YAML schema with the config min/max instead.

In theory this could be used to put future password complexity settings as well.


amosfolz:
Can you change the config to be under the same subkey is in pending changes here: https://github.com/userfrosting/UserFrosting/pull/974/files#diff-2b7f59d74c443f25aace0c67755550d1R118

lcharette:
> Can you change the config to be under the same subkey is in pending changes here: https://github.com/userfrosting/UserFrosting/pull/974/files#diff-2b7f59d74c443f25aace0c67755550d1R118

Or maybe `site => password => security` ?

avsdev-cw:
The latter would probably be better from a future merging point of view

PR #1000, opened 2019-06-13
Add missing locale keys (still need translated)

amosfolz:
Generated missing locale keys for all languages using PR #992 `locale:fix-keys` command.

The empty key values were populated using en_US and can be identified via `locales:missing-values` command. 

(also wanted to check these files will pass style-ci check)


amosfolz:
This branch/PR has become very dirty. Changes have been moved to PR #1006

PR #1006, opened 2019-07-01
Update Locale File Keys

amosfolz:
Generated missing locale keys for all languages using PR #992 locale:fix-keys command.

The empty key values were populated using en_US and can be identified via locales:missing-values command.

(also wanted to check these files will pass style-ci check)

I also had to go through and modify some files manually to fix keys that were present in other locales but not in en_US. Some of these were due to keys not being in correct sub keys or improper formatting.


app/sprinkles/account/locale/ar/validate.php
```diff
@@ -15,7 +15,8 @@
  * @author Abdullah Seba
  */
 return [
-    'VALIDATE' => [
-        'PASSWORD_MISMATCH' => 'يجب أن تكون كلمة المرور وكلمة المرور التأكيدية نفس',
```
lcharette: Not sure why the indentation changed here... Should be 4 spaces. Also not sure why php-cs-fixer can't fix that and StyleCI allows that...

lcharette: Fixed

lcharette:
Would it be easier for translators if we comment all untranslated keys? I know it won't work with the upcoming Bakery commands, but it would make it easier to spot what needs translation and actually use the English fallback instead of having outdated English strings in whatever locale.

amosfolz:
> 
> 
> Would it be easier for translators if we comment all untranslated keys? I know it won't work with the upcoming Bakery commands, but it would make it easier to spot what needs translation and actually use the English fallback instead of having outdated English strings in whatever locale.

I can do that if you think it's best. 

I do have a table with the untranslated keys that I was going to put into a new issue for translators to reference.

amosfolz:
I commented out the keys in recent commits.

lcharette:
```PHP Parse error:  syntax error, unexpected ';', expecting ']' in /home/travis/build/userfrosting/UserFrosting/app/sprinkles/core/config/default.php on line 408

https://travis-ci.org/userfrosting/UserFrosting/jobs/571120012#L881

amosfolz:
> 
> 
> ```
> PHP Parse error:  syntax error, unexpected ';', expecting ']' in /home/travis/build/userfrosting/UserFrosting/app/sprinkles/core/config/default.php on line 408
> ```
> 
> https://travis-ci.org/userfrosting/UserFrosting/jobs/571120012#L881

I'm not even sure where this change came from?

lcharette:
Something barfed 🤔 
https://travis-ci.org/userfrosting/UserFrosting/jobs/571375743#L881

amosfolz:
I don't even understand what happened here... 
Missing closing bracket https://github.com/amosfolz/UserFrosting/blame/hotfix-update-locale-files/app/sprinkles/core/config/default.php#L105-L111 ?

amosfolz:
Also, php-cs-fixer is trying to do weird things like change comments now:
image

lcharette:
I think it should be safe to merge now? PHP-CS-Fixer didn't brought up any issue on my side?

amosfolz:
> 
> 
> I think it should be safe to merge now? PHP-CS-Fixer didn't brought up any issue on my side?

I think it is good to go.

Issue #1002, opened 2019-06-24
php bakery migrate:rollback missing class errors


amosfolz:
Trying to run `php bakery migrate:rollback` will fail if there is a missing migration class, even if it is not in the last batch that was ran. 

![image](https://user-images.githubusercontent.com/33728190/60052886-7730ca80-96a4-11e9-8f3b-23a088720fca.png)

![image](https://user-images.githubusercontent.com/33728190/60052900-8283f600-96a4-11e9-8527-0b9ebd4d3a75.png)

It seems it shouldn't matter if a migration class is missing unless that class is part of the last batch ran, since `php bakery migrate:rollback` just reverses the most recent migration. Unless `migrations` != `batches`?

lcharette:
Can you pinpoint where this exception is thrown ?

amosfolz:
> 
> 
> Can you pinpoint where this exception is thrown ?

It seems to be from this [line](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Database/Migrator/MigrationDependencyAnalyser.php#L213)

Rather than trying to modify anything in any of the `Migrator` classes, a simpler solution might be to ask the user if they want to drop the "stale" record from the database table.

lcharette:
That might be a good idea... `migrate:clean` ?

amosfolz:
```
#0  UserFrosting\Sprinkle\Core\Database\Migrator\MigrationDependencyAnalyser->getMigrationDependencies() called at [/home/vagrant/userfrosting/app/sprinkles/core/src/Database/Migrator/MigrationRollbackDependencyAnalyser.php:66]
#1  UserFrosting\Sprinkle\Core\Database\Migrator\MigrationRollbackDependencyAnalyser->validateClassDependencies() called at [/home/vagrant/userfrosting/app/sprinkles/core/src/Database/Migrator/MigrationDependencyAnalyser.php:78]
#2  UserFrosting\Sprinkle\Core\Database\Migrator\MigrationDependencyAnalyser->analyse() called at [/home/vagrant/userfrosting/app/sprinkles/core/src/Database/Migrator/MigrationDependencyAnalyser.php:161]
#3  UserFrosting\Sprinkle\Core\Database\Migrator\MigrationDependencyAnalyser->getUnfulfillable() called at [/home/vagrant/userfrosting/app/sprinkles/core/src/Database/Migrator/Migrator.php:327]
[...]

PR #1007, opened 2019-07-02
migrate:clean bakery command

amosfolz:
This command removes stale migrations from the database. These missing migration classes prevent other `migrate:*` commands from running until the missing class files are either added back to the file system or  manually dropped from the `migrations` table. 

I also added a note about the new command when "Unable to find migration..." error messages are given in two of the `Migrator` classes. 

Issue #1002

### DIFF
--- a/app/sprinkles/core/src/Bakery/MigrateCleanCommand.php
+++ b/app/sprinkles/core/src/Bakery/MigrateCleanCommand.php
@@ -31,7 +31,9 @@ class MigrateCleanCommand extends MigrateCommand
     {
         $this->setName('migrate:clean')
              ->setDescription('Remove stale migrations from the database.')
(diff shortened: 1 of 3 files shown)


amosfolz:
It looks like the `5.4` builds are failing

lcharette:
Looks good. I'll test the actual command when I get the chance, but it could be quite useful in development. Might just add a `help` part where you explain what a stale migration is and why it can be dangerous to removed them (failed dependencies). A `force` option could be useful in this case to by-pass the confirmation.

amosfolz:
> 
> 
> Looks good. I'll test the actual command when I get the chance, but it could be quite useful in development. Might just add a `help` part where you explain what a stale migration is and why it can be dangerous to removed them (failed dependencies). A `force` option could be useful in this case to by-pass the confirmation.

Added help message and `force` option in recent commit.

PR #1017, opened 2019-07-11
Set password manually when creating new user

amosfolz:
I spread #fakenews about https://github.com/userfrosting/UserFrosting/pull/764 and this feature was not actually implemented (sorry).

I forked the branch and resolved merge conflicts. Everything seems to checkout fine.


app/sprinkles/admin/src/Controller/UserController.php
```diff
@@ -133,8 +133,12 @@ public function create(Request $request, Response $response, $args)
         }
 
         $data['flag_verified'] = 1;
-        // Set password as empty on initial creation.  We will then send email so new user can set it themselves via a verification token
-        $data['password'] = '';
+        if (!isset($data['value'])) {
```
lcharette: Is there a reason the field is called `value` and not `password` ?

Issue #713, opened 2017-05-24
Icon picker for user and group form


lcharette:
Wherever you can specify an icon in the UI (Group & user form), we should have an icon picker for non dev people to use. 

See : https://itsjavi.com/fontawesome-iconpicker/

amosfolz:
I should mention I started working on this but it relies on #870.

PR #1019, opened 2019-07-20
Feature: Font Awesome 5 Icon Picker

amosfolz:
Adds a drop-down menu to allow admin to select group icon from a list of available icons. 

Ref: https://github.com/userfrosting/UserFrosting/issues/713


lcharette:
The result is amazing. Great new feature, very useful ! Thanks !

PR #1028, opened 2019-08-31
Update Arabic Locales

abdullahseba:


app/sprinkles/core/locale/ar/messages.php
```diff
@@ -57,11 +57,11 @@
         '@TRANSLATION' => 'سياسة الخصوصية',
         'DESCRIPTION'  => 'تحدد سياسة الخصوصية لدينا نوع المعلومات التي نجمعها منك وكيفية استخدامها.',
     ],
-    // 'SLUG'                     => 'Slug',
+    'SLUG'                     => 'سبيكة',
     'SLUG_CONDITION'           => 'Slug/الظروف',
```
amosfolz: I wasn't sure if there is a comparable word for `slug`. If so, should this line be updated?

abdullahseba:
I didn't know of that word either as it happens.  I was expecting a slug:
image
But the word I used means slug as in metal alloy so I could update it I guess.  I'm not too sure how an actual Arab person would understand it.

amosfolz:
> 
> 
> I didn't know of that word either as it happens. I was expecting a slug:
> image
> But the word I used means slug as in metal alloy so I could update it I guess. I'm not too sure how an actual Arab person would understand it.

I'm not sure it needed to be changed, just wanted to make sure you saw that one. 

Everything else looks great. Thanks for helping.

PR #1030, opened 2019-09-02
Fix typos in user creation with initial password

rwos:
Not sure if I picked the right base branch, let me know :)

This fixes some (I guess) typos in https://github.com/userfrosting/UserFrosting/pull/1017 such that it actually works (on my machine™️).

issue: https://github.com/userfrosting/UserFrosting/pull/764

### DIFF
--- a/app/sprinkles/admin/src/Controller/UserController.php
+++ b/app/sprinkles/admin/src/Controller/UserController.php
@@ -167,7 +167,7 @@ class UserController extends SimpleController
             $passwordRequest = $this->ci->repoPasswordReset->create($user, $config['password_reset.timeouts.create']);
 
             // If the password_mode is manual, do not send an email to set it. Else, send the email.
-            if (!isset($data['value'])) {
+            if ($data['password'] === '') {
                 // Create and send welcome email with password set link
                 $message = new TwigMailMessage($this->ci->view, 'mail/password-create.html.twig');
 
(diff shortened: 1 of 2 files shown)


amosfolz:
Good catch. Looks like I missed that line when I updated it in https://github.com/userfrosting/UserFrosting/pull/1017/commits/acf4d600b535e08fad462c70cbebc627ba0754c8.

rwos:
with the test failure, I'm not so sure anymore if this was actually the correct fix...

amosfolz:
I think the test may need to be adjusted, as before it assumed there was never a password field to handle.

lcharette:
https://github.com/userfrosting/UserFrosting/blob/1f3edc0c92751cb1ec5490ca94b6de6aa1ed5994/app/sprinkles/admin/src/Controller/UserController.php#L789-L790

app/sprinkles/admin/src/Controller/UserController.php
```diff
@@ -167,7 +167,7 @@ public function create(Request $request, Response $response, $args)
             $passwordRequest = $this->ci->repoPasswordReset->create($user, $config['password_reset.timeouts.create']);
 
             // If the password_mode is manual, do not send an email to set it. Else, send the email.
-            if (!isset($data['value'])) {
+            if ($data['password'] === '') {
```
rwos: The test failure was actually due to the `isset` here - it's set to empty string if not set some lines earlier.

amosfolz:
> 
> 
> Not sure if I picked the right base branch, let me know :)
> 
> This fixes some (I guess) typos in #1017 such that it actually works (on my machine™️).
> 
> issue: #764

Just a note, it would be helpful to include some details about what's actually broken for others to be able to tell what exactly this is fixing. :-) "Such that it actually works" is not very helpful.

rwos:
@amosfolz sure, sorry about that - for the record, the problem was that the frontend validation wouldn't allow entering a password, because it always failed the check against the password confirmation field.

PR #1033, opened 2019-09-11
Fix typos in form.fields.hidden

phillmac:
Fix typos: missing `form.`  in `form.fields.hidden` for icon and description

Noticed that these two fields were not being hidden despite including them in `form.fields.hidden` in the controller.

### DIFF
--- a/app/sprinkles/admin/templates/forms/group.html.twig
+++ b/app/sprinkles/admin/templates/forms/group.html.twig
@@ -33,7 +33,7 @@
                 </div>
             </div>
             {% endif %}
-            {% if 'icon' not in fields.hidden %}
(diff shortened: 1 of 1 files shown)


amosfolz:
`fields.hidden` seems to be in a few other places as well: https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/admin/templates/forms/role.html.twig#L35

Are we sure that wasn't intentional?

lcharette:
Unless you're in a for loop, should start with form : https://github.com/userfrosting/UserFrosting/blob/79d87fd398e8f8992a56bcac6d8dd382aaaa93a4/app/sprinkles/admin/src/Controller/RoleController.php#L427

PR #1034, opened 2019-09-16
Fix edit password schema field name

phillmac:
The input name in the rendered modal is `password` instead of `value`
See https://github.com/userfrosting/UserFrosting/blob/a8e6711fd5afd0572fc1581950f161b9581a8ffd/app/sprinkles/admin/templates/forms/partials/user-set-password.html.twig#L25

### DIFF
--- a/app/sprinkles/admin/schema/requests/user/edit-password.yaml
+++ b/app/sprinkles/admin/schema/requests/user/edit-password.yaml
@@ -1,5 +1,5 @@
 ---
-value:
+password:
   validators:
     required:
       domain: client
@@ -19,7 +19,7 @@ passwordc:
       message: VALIDATE.REQUIRED
     matches:
       domain: client
-      field: value
+      field: password
       label: "&PASSWORD.CONFIRM"
       message: VALIDATE.PASSWORD_MISMATCH
     length:


lcharette:
I think the whole confusion here might comes from the `upateField` controller method. This method can accept any fields, one or more, and update it, and at some point used `value` for something... 

I think there might also be something wrong in the controller, as `$fieldName` is defined twice. I'll have to check out what cause that :  

https://github.com/userfrosting/UserFrosting/blob/a8e6711fd5afd0572fc1581950f161b9581a8ffd/app/sprinkles/admin/src/Controller/UserController.php#L1288

https://github.com/userfrosting/UserFrosting/blob/a8e6711fd5afd0572fc1581950f161b9581a8ffd/app/sprinkles/admin/src/Controller/UserController.php#L1324

phillmac:
The set password schema is for generating the client side validation rules as the `domain` is set to `client` so it should never be loaded in a controller

lcharette:
I've merged this, and along with #1038, made other improvement in https://github.com/userfrosting/UserFrosting/commit/29bfdd41189315a51bfc48f321fd8b416b696a68. Should be good to go now.

PR #1035, opened 2019-09-24
Hotfix sr rs

zbigcheese:
Serbian translations for the base sprinkles


lcharette:
Can you add `sr_RS` to the default config file ?

amosfolz:
I'm not sure where in the order this one should go... Config says to list in order according to https://en.wikipedia.org/wiki/List_of_languages_by_total_number_of_speakers

lcharette:
At this point, Alphabetically is probably easier for people to find in a dropdown. @alexweissman ?

lcharette:
Actually, with the planned changes in 4.4 and 5.0, the list will probably be alphabetical as it will be dynamically loaded or user defined in the config...

PR #1038, opened 2019-10-01
Fix direct password edit

rwos:
Bug: editing a users password does not work, it always ends up as invalid because "the passwords don't match".

This fixes that by making the name of the password field flexible:
* 'password' on the user creation form
* 'value' on the edit password form (since that goes through admin's `UserController::updateField`)

this replaces https://github.com/userfrosting/UserFrosting/pull/1034
see also https://github.com/userfrosting/UserFrosting/pull/1030

I now see that the whole thing could've also been fixed by calling the password field `"value"` (and reverting the above PR), but that looks a bit weird IMHO. But let me know what you think!

### DIFF
--- a/app/sprinkles/admin/src/Controller/UserController.php
+++ b/app/sprinkles/admin/src/Controller/UserController.php
@@ -77,6 +77,10 @@ class UserController extends SimpleController
 
         // Load the request schema
         $schema = new RequestSchema('schema://requests/user/create.yaml');
+        $schema->set('password.validators.length.min', $config['site.password.length.min']);
+        $schema->set('password.validators.length.max', $config['site.password.length.max']);
+        $schema->set('passwordc.validators.length.min', $config['site.password.length.min']);
+        $schema->set('passwordc.validators.length.max', $config['site.password.length.max']);
 
         // Whitelist and set parameter defaults
         $transformer = new RequestDataTransformer($schema);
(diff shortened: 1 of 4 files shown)


amosfolz:
@rwos Have you tested to make sure this is compatible with updating existing users password?

rwos:
@amosfolz that's exactly the thing that this fixes. But I would appreciate someone other than me checking this again with v4.3.1 (setting a manual password on user creation works, updating a user's password is broken) and this PR (both work). I'm reasonably sure this works, but I was reasonably sure in https://github.com/userfrosting/UserFrosting/pull/1030 too :-)

n0use:
I just tried this on 4.3.1 and it does fix CHANGING a users password, which is good, but I get the same "Your password and confirmation password must match." error when trying to CREATE a user with a password set.

amosfolz:
I think this still needs some work. For example, validation messages appear not to match anymore:

image

lcharette:
https://github.com/userfrosting/UserFrosting/pull/1034#issuecomment-531794028

The "single controller method accepting any field" thing could be phased out for password if this is too much of an issue.

rwos:
Alright - validation should be fixed now, and it now also uses the `site.password.length` config everywhere (edit password + user create), as it should.

lcharette:
So... Which one between this (#1038) and #1034 actually fix this issue?

amosfolz:
This is the one I tested - everything worked besides the validation message. If that is fixed, this should be merged and the other closed.

lcharette:
I've merged this, and with #1034, made other improvement in https://github.com/userfrosting/UserFrosting/commit/29bfdd41189315a51bfc48f321fd8b416b696a68. Should be good to go now.

I change the behavior of `UserController::updateField`.  Instead of using a common `value` key for all fields, the key is now the field name itself. 

For example, when updating `flag_enable`, the data should be `flag_enabled => 1` instead of `value => 1`. So now `edit-password.yaml` [will have the proper `password` value in it](https://github.com/userfrosting/UserFrosting/pull/1034/files). 

Could introduce a minor Breaking change if someone uses that class in it's own sprinkle. Will add a note in the changelog.

Issue #1027, opened 2019-08-30
Locale translation files need translated


amosfolz:
Missing translation keys were added in `4.3` but we still need translators to assist with translations. 

The keys that need translated have been commented out in each locale file. 

- [x] ar (Arabic)
- [x] de_DE (German)
- [x] el (Greek)
- [x] fa (Persian)
- [x] fr_FR (French)
- [ ] it_IT (Italian)
- [ ] pt_PT (Portuguese)
- [x] ru_RU (Russian) 
- [x] th_TH (Thai)
- [ ] tr (Turkish)
- [ ] zh_CN (Chinese)

Le-Morri:
Hey, I'm happy to do the de_DE locale. Let me know if you're still interested in it

amosfolz:
@Le-Morri They have not been updated, please do. They keys are commented out that need updated.

Le-Morri:
@amosfolz thanks for the hint. to which branch should I open the PR? as for the arabic to hotfix?

lcharette:
> which branch should I open the PR?

Hotfix

uatthaphon:
Hi, I want to take th_TH (Thai) if nobody works on it (^v^).

nikolaos-spyratos:
I can take on Greek.

lcharette:
@nikolaos-spyratos [Might still need updating](https://github.com/userfrosting/UserFrosting/blob/hotfix/app/sprinkles/account/locale/el/messages.php#L65) :)

arminkhoshbin:
@amosfolz #1045 for the missing Persian translations.

alannaidon:
PT_BR isn't required at the moment? I could help with that.

lcharette:
@alannaidon There's no `PT_BR` locale yet, but you're welcome to add it if you want :)

nina-py:
@amosfolz, just a quick note to say that the ru_RU locale appears to be 100% complete and should be ticked off your list.

amosfolz:
> 
> 
> @amosfolz, just a quick note to say that the ru_RU locale appears to be 100% complete and should be ticked off your list.

Thanks for confirmation

josemachado94:
#1178  -  pt_PT review

lcharette:
A new issue will be opened for UF5

PR #1039, opened 2019-10-01
German localisation

Le-Morri:
As discussed in issue #1027 the updated german (de_DE) locals


amosfolz:
Looks good to me.

Thank you.

PR #1042, opened 2019-10-09
Add missing HEADER_MESSAGE_ROOT key, updated some English to Greek

nikolaos-spyratos:
@lcharette Kindly pointed out that there was still a missing locale key, so I added that.

Also modified two other keys to be less Greek-lish.

Let me know if there's anything missing.


amosfolz:
Is there anything more appropriate for `captcha`, `email`, or `slug`?
https://github.com/nikolaos-spyratos/UserFrosting/blob/feature/greek-locale-additions/app/sprinkles/core/locale/el/messages.php#L15

nikolaos-spyratos:
On my old branch where I did my own version of the translations I also couldn't find a good translation for `slug`. In Greek, slug is translated as "naked snail" and the language isn't always as malleable with using words for new meanings.
`email` is colloquially used in Greece instead of the transliteration (which basically means "electronic post").
`captcha` is more of a pronoun on its own and Greek maintains foreign terms like that.

lcharette:
FYI, for the French locale I used the equivalent of "token" instead of slug. Because the only translation in French would also be "naked snail" :P

lcharette:
Merged ! Thanks! 

Ref #1027

PR #1062, opened 2020-01-29
Improve ordering by activity date

ktecho:
Some activity logs can occur at the same second, so ordering just by date can show the results unordered. As per @lcharette suggestion, we can order by `occurred_at` AND `id`. That should gives better results.

### DIFF
--- a/app/sprinkles/admin/src/Sprunje/ActivitySprunje.php
+++ b/app/sprinkles/admin/src/Sprunje/ActivitySprunje.php
@@ -68,7 +68,7 @@ class ActivitySprunje extends Sprunje
 
         return $this;
     }
-    
+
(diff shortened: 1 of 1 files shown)


lcharette:
The style issue will need to be updated first : https://github.styleci.io/analyses/OML5Lk?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

ktecho:
@lcharette I re-added the whitespace but it's giving two fails. Do you know what could be happening?

lcharette:
Nan, you need to remove the whitespace. 

FYI, Codecov failure can be ignored

ktecho:
@lcharette Should be done now

PR #1065, opened 2020-02-08
Adds GitHub actions for testing & Windows

### LINKED ISSUE #933: Setup Windows & SQL Server Testing Environment
Support for SQL Server on Windows seams broken with UserFrosting 4.2, at least at the testing level. 

Current AppVeyor error message : 
```
SQLSTATE[23000]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Cannot insert explicit value for identity column in table 'users' when IDENTITY_INSERT is set to OFF. (SQL: insert into [users] ([user_name], [first_name], [last_name], [email], [locale], [flag_verified], [flag_enabled], [password], [id], [updated_at], [created_at]) values (Cade, Grover, Collin, yoshiko04@gmail.com, en_US, 1, 1, .h56897nO%QzBfvFgQK.6, 2019-02-01 14:39:50.000, 2019-02-01 14:39:50.000, ?))
```
https://ci.appveyor.com/project/userfrosting/userfrosting/build/job/7j5xhwb19tthe1w2#L2959 

Everything I found so far about this issue and Eloquent is related to the DB Transaction system and Eloquent 5.4 or prior. This leads me to believe this issue might be fixed in newer versions of Laravel/Eloquent. 

I won't merge the [`develop-windows`](https://github.com/userfrosting/UserFrosting/tree/develop-windows) branch just yet, as it contains only the AppVeyor config anyway. This should be monitored and reassessed once #907 is done.

Right now the path issue seen before on Windows should be fixed with UserFrosting/UniformResourceLocator 4.2.2.

lcharette:
Note this doesn't work in it's current state. MySQL and SQL Server for Windows isn't supported but GitHub Action yet. 

On Linux side, it could work if MySQL config can be fixed. But it's still less than Travis, as pgsql is not supported on GitHub Action yet. On the other side, GitHub Action is wayyy faster. So Travis is still better as long as Windows can work here. 
 
Also, see comment in #933:
> GitHub Actions looks promising, as they are fully documented, new and sparkling, but it doesn't support SQL Server yet. PHP support can be enabled using [setup-php](https://github.com/shivammathur/setup-php/tree/1.8.2), but it still doesn't support MySQL and Redis on Windows (throws the `Container operations are only supported on Linux runners` otherwise). 

Alternatively, see #1064 for App Veyor implementation which could work for Windows.


lcharette:
Replaced with #1132

PR #1073, opened 2020-03-27
Sprunje Bug fix and Docker updates

### LINKED ISSUE #1068: Sprunje applyTransformations method not returning the $collections object
The Sprunje is not utilizing the return value from `applyTransformations` method in the `Sprunje` class.

https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Sprunje/Sprunje.php#L314 
This should be  `$collection = $this->applyTransformations($collection);`  instead.
There is another instance of this in the `Sprunje.php` file 
https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Sprunje/Sprunje.php#L245

ssnukala:
1. Fixes #1068 
2. Docker Changes 
 - Added configurations for Node
 - Added reverse-proxy so the docker image is accessible using http://userfrosting.localhost

### DIFF
--- a/app/sprinkles/core/src/Sprunje/Sprunje.php
+++ b/app/sprinkles/core/src/Sprunje/Sprunje.php
@@ -195,7 +195,7 @@ abstract class Sprunje
             $response = $response->withAddedHeader('Content-Type', 'text/csv; charset=utf-8');
 
             return $response->write($result);
-            // Default to JSON
+        // Default to JSON
         } else {
             $result = $this->getArray();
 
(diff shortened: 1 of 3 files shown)


lcharette:
Could you split this PR in two? Sprunje fix can be merged, but not so sure about Docker, it needs more approval. Thanks !

Silic0nS0ldier:
Docker I would do in a separate PR. Docker change could break peoples workflows, a closer inspection is needed which would hold up the Sprunje fix.

ssnukala:
Ok I will split it, I should have done this separately in the first place - sorry

Issue #1067, opened 2020-03-05
Add Bakery options for: smtp:<port>, secure:<true|false>, auth:<true|false>


waldenn:
In order to make the email functionality work on my SMTP server (which forwards emails to a secure ISP relay), I had to make these changes in the userfrosting/app/sprinkles/core/config/default.php:

>   'port' => 25,
>   'auth' => false,
>   'secure' => 'false'

Please consider adding these options to the setup options (and documentation).
I had to create a PHPMailer application to debug my SMTP connection problem, hopefully other people can avoid that with these settings.

Thanks for a great user management framework!

gyaaniguy:
I am going for a PR. How should this be ideally handled ?

lcharette:
@gyaaniguy Thanks!

I see three things that needs to be done : 

1. Add new ENV variables to `.env.example`
1. Implement said variables to the config file
1. Add the new options to the [SetupSmtpCommand  Bakery command](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Bakery/SetupSmtpCommand.php#L26).

mikkoaf:
I was looking through up-for-grabs issues. 

I found out that the linked, closed PR https://github.com/userfrosting/UserFrosting/pull/1077 was merged to master in https://github.com/userfrosting/UserFrosting/commit/82a1706c902f37fb690ae7b2019bb5822f0106e2 and is included on the release https://github.com/userfrosting/UserFrosting/releases/tag/v4.5.0

Can this issue be closed or is there still work to be done?

PR #1077, opened 2020-04-08
More SMTP options in config #1067

gyaaniguy:
Added SMTP port, SMTP authentication type , and SMTP security protocol to config file. So that it can be set from .env file.
I did not get a chance to test it. Apologies if I did something incorrect.


app/sprinkles/core/config/default.php
```diff
@@ -232,9 +232,9 @@
     'mail'    => [
         'mailer'          => 'smtp', // Set to one of 'smtp', 'mail', 'qmail', 'sendmail'
         'host'            => getenv('SMTP_HOST') ?: null,
-        'port'            => 587,
-        'auth'            => true,
-        'secure'          => 'tls', // Enable TLS encryption. Set to `tls`, `ssl` or `false` (to disabled)
+        'port'            => getenv('SMTP_PORT') ?: null,
+        'auth'            => getenv('SMTP_AUTH') ?: null,
+        'secure'          => getenv('SMTP_SECURE') ?: null, // Enable TLS encryption. Set to `tls`, `ssl` or `false` (to disabled)
```
lcharette, line 237: Hum... This would create a breaking change for people without those settings already in their env, which push this change to UF 4.5. I wonder if there's a way to keep the default here, and overwrite in the env?

app/sprinkles/core/src/Bakery/SetupSmtpCommand.php
```diff
@@ -169,11 +179,17 @@ protected function askForSmtp(InputInterface $input)
             // Use custom validator to accept empty password
             return $password;
         });
+        $smtpPort = ($input->getOption('smtp_port')) ?: $this->io->ask('SMTP Server Port', 587);
+        $smtpAuth = ($input->getOption('smtp_auth')) ?: $this->io->ask('SMTP Server Authentication', 'true');
+        $smtpSecure = ($input->getOption('smtp_secure')) ?: $this->io->ask('SMTP Server Security type', 'tls');
```
lcharette, line 184: A yes/no question or choices could probably better here.

alexweissman:
I would maybe try to make it consistent with Laravel's environment variables? See https://github.com/laravel/laravel/blob/master/.env.example

lcharette:
I manually merged in 82a1706c as I wasn't able to update the original branch. I manually fixed the comments and conflict at the same time.

Issue #1076, opened 2020-04-06
Why are successful sign-in attempts also throttled?


rwos:
Hi!

I'm having some problems with the `sign_in_attempt` throttler also counting successful sign-ins. In my case, it's mostly integration tests that run into this, but depending on how people set up that throttler, it could also affect "normal" users.

The [code for that](https://github.com/userfrosting/UserFrosting/blob/c62790e99e8c68d5f48849bf561a8ed293c0a9f2/app/sprinkles/account/src/Controller/AccountController.php#L400) looks intentional to me, so I guess I'm mostly asking for the rationale behind this. IMHO, it would make more sense to only throttle *failing* logins. I could do a PR for that, but wanted to open an issue first :)

lcharette:
Ping @alexweissman

alexweissman:
I think the original assumption was that any rapid series of login attempts, failed or successful, would be considered suspicious behavior and therefore get throttled. However, you make a good point that this doesn't account for automated testing, and there are likely some legitimate cases for making multiple _successful_ login attempts in a short period of time.

Based on what I'm seeing in OWASP (for example, https://owasp.org/www-community/controls/Blocking_Brute_Force_Attacks) there isn't really a case for throttling successful login attempts. So, I would accept a PR to change the current behavior.

lcharette:
Side note, for automated testing, unless testing the throttle itself, it should probably be mocked and ignored.

Silic0nS0ldier:
> I think the original assumption was that any rapid series of login attempts, failed or successful, would be considered suspicious behavior and therefore get throttled.
That exact behaviour is what would be seen in a credential stuffing attack. Making automated testing exempt from the throttle will need some care to ensure we don't introduce an exploitable behaviour.

alexweissman:
True, but credential stuffing would involve a different set of credentials with each attempt, no? Seems like we would need more complex logic than what our current implementation currently supports.

PR #1082, opened 2020-04-21
sign_in_attempt throttler: don't count successful logins

rwos:
for https://github.com/userfrosting/UserFrosting/issues/1076

This only changes how attempts are counted, not how the blocking works.

If you got the password wrong often enough to be above the throttle limit, you still can't login, even if you have the correct password now (as that would defeat the whole point).

The big upside is that one can now configure `sign_in_attempts` to be *much* more strict (one could even start blocking after the first attempt).

Tested via throttler mock - let me know if you have better ideas :)

### DIFF
--- a/app/sprinkles/account/src/Controller/AccountController.php
+++ b/app/sprinkles/account/src/Controller/AccountController.php
@@ -396,13 +396,11 @@ class AccountController extends SimpleController
             return $response->withJson([], 429);
         }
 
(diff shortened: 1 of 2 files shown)


Silic0nS0ldier:
Haven't checked the code, but the logic you've explained is sound. LGTM spec wise.

lcharette:
Sounds great, good job, thanks !

PR #1083, opened 2020-04-23
4.4.1 Release

lcharette:


app/sprinkles/core/src/Twig/CoreExtension.php
```diff
@@ -49,20 +53,20 @@ public function getName()
     /**
      * Adds Twig functions `getAlerts` and `translate`.
      *
-     * @return array[\Twig_SimpleFunction]
+     * @return array[TwigFunction]
```
Silic0nS0ldier, line 56: Fairly sure the correct type annotation is `TwigFunction[]` here. Current format indicates an array of mixed values where keys are of type `TwigFunction`.

app/sprinkles/core/src/Twig/CoreExtension.php
```diff
@@ -73,7 +77,7 @@ public function getFunctions()
     /**
      * Adds Twig filters `unescape`.
      *
-     * @return array[\Twig_SimpleFilter]
+     * @return array[TwigFilter]
```
Silic0nS0ldier, line 80: As before. `TwigFilter[]`

app/sprinkles/core/src/Twig/CoreExtension.php
lcharette: I'll have to look at the doc, but both seams accepted by my IDE ¯\_(ツ)_/¯

Silic0nS0ldier: IDE type analysers are rather inconsistent here, speaking from experience (3 different VSCode PHP intellisense plugins handled array syntax differently, the spec way at least was consistent).
https://docs.phpdoc.org/latest/guides/types.html#arrays

Silic0nS0ldier: Just to add, one of the plugins treated the above syntax like it indicates the array key type. Which frankly is rather neat and would be nice to see become a standard.

lcharette: I've used `array<key,value>` in my Sprinkle. Works with VSCode Intelephense.

lcharette: Anyway, I guess it can wait for later. More in those class that need to be fixed...

lcharette: https://github.com/php-fig/fig-standards/blob/211063eed7f4d9b4514b728d7b1810d9b3379dd1/proposed/phpdoc.md#collections

lcharette: (Apparently that got rejected somehow...)

Silic0nS0ldier: Fair enough.

Re the proposal, kinda glad about the outcome. I'd rather that syntax be reserved for generics, a much more impactful feature.

app/sprinkles/core/src/Twig/CoreExtension.php
Silic0nS0ldier: As per above thread, this is to be followed up at a later time.

PR #1089, opened 2020-04-26
Moving things around in preparation for 5.0

lcharette:
In preparation for the skeleton repo feature of 5.0, I propose we move much from System into Core right now to make things easier to start 5.0, and ease the upgrade process.

Changes in this PR : 
- Moved constants from `app/define.php` to `app/sprinkles/core/define.php`.
- Moved some requirements from `/composer.json` to `app/sprinkles/core/composer.json`.
- Moved `UserFrosting\Tests\TestCase` to `UserFrosting\Sprinkle\Core\Tests\TestCase`.
- Moved `UserFrosting\System\Sprinkle\Sprinkle` to `UserFrosting\Sprinkle\Core\Sprinkle\Sprinkle`.
- Moved `UserFrosting\System\Facade` to `UserFrosting\Sprinkle\Core\Facade`.
- Moved `UserFrosting\System\SlimAppEvent` to `UserFrosting\Sprinkle\Core\SlimAppEvent`.
- Moved `UserFrosting\System\Sprinkle\SprinkleManager` to `UserFrosting\Sprinkle\Core\Sprinkle\SprinkleManager`.
- Moved `UserFrosting\System\UserFrosting` to `UserFrosting\Sprinkle\Core\UserFrosting`.
- Moved `*Command` Bakery class to `Command/*` namespace.
- Replaced `UserFrosting\System\Bakery\BaseCommand` with `UserFrosting\Sprinkle\Core\Bakery\AbstractCommand`.

There's still things that could be improved here. Most code in `UserFrosting\Sprinkle\Core\UserFrosting` could be moved to `UserFrosting\Sprinkle\Core\Core`, including improvements to how the services are loaded. 

For now, I moved services from the deleted `app/system/ServicesProvider.php` into `UserFrosting\Sprinkle\Core\UserFrosting`, but I think it could be further improved. More precisely, I think the order the sprinkle are setup could be improved. Currently, it setup all the location, then load all services, which makes it difficult, as the locator service needs to be loaded first, then the locations added, then the other services can be loaded. 

Could probably be `Sprinkle A -> services, paths; Sprinkle B -> Services, path, etc.` instead. 

Furthermore, I think we should move bakery commands from "magic discovery" to "deliberate registration". I could start with this release to.


lcharette:
Ok, so after a struggle with case sensitive filesystem on Travis, and case insensitive filesystem on Vagrant , I think the SprinkleManager / UserFrosting main class are probably 90% finished now compared to the previous message. 

Just a note, some methods regarding sprinkle names are not case insensitive anymore. A note will need to be added to the learn doc just to be sure, but the idea I went with is the name in `sprinkle.json` needs to be the same case as the directory (so `app/Sprinkles/PasTries/` needs to be as `PasTries` in `sprinkle.json`). This won't matter when / if sprinkles are loaded by composer, because you can specify the folder name in the Sprinkle composer file.

Some todo's still left (and will be done tonight probably), but logic should not change much for review.

lcharette:
Ok, one thing I just thought of... By moving the TestCase, it will have the most impact on Sprinkles. This means my sprinkles that currently support UF 4.3 and 4.4 won't be able to also support 4.5 because of the tests.

I think a `class_alias` or deprecation for the TestCase would make it easier for sprinkles for this release...

Silic0nS0ldier:
I'm all for increasing the amount of code which is deliberately registered. Aside from simplifying logic, it is likely to present better performance once JIT arrives in PHP 8 (reflection will gain benefits from the JIT, however static code will benefit more thanks to the complier not having to "rush" the job).

What exactly is going to be the responsibility of `Core` in UF 5? Delivery of framework updates? If so, then I believe we'll need a different approach where a lightweight bootstrapper package handles wiring everything up, and the core sprinkle brings in the default experience existing users are familiar with. This would avoid muddying the definition of a Sprinkle (a mechanism for extensibility).

lcharette:
Yes, and the relation with the other supporting packages needs to be taken into consideration...

On one end, their structure make them easy to test, but on the other hand very hard to manage dependency wise. Core is the opposite...

As for the défaut experience, this definitely needs to be handled by a new ‘site’ sprinkle.

lcharette:
N.B.: I've marked this PR back to draft, as I need to think a bit more about this. 

While moving the Bakery command to the right namespace makes sense, I don't think moving the `TestCase` does not at this point. Especially if we're not sure this will be their final namespace... (TestCase could end up in a `Test` sprinkle... or dependencies)

lcharette:
I'll close this PR and start a new one, I need to rework theses changes.

PR #1109, opened 2020-10-05
Check for null key in Unique::buildTertiaryDictionary

alexweissman:
Similar to what's done in Unique::buildDictionary, we need to allow for null values in the third column, in which case there is no tertiary model to be added for this row.


lcharette:
Since the base got messed up, I merged it manually in Hotfix: ac463542803531c47b7547a873cdc56a3126a12d

Issue #820, opened 2017-12-05
Email reverification after change


splitt3r:
The UF config allows us to turn on user verification (`site.registration.require_email_verification`) but doesn´t reverify the email after the user changes it. This option is pretty useless if the user can use a verified email to register and change it afterwards to an unverified one (without any history). When the user tries to change the email address the system should force the user to verify the new email again. In a lot of scenarios you want to be able to reach your users / verify that they have at least a real / working email address. I know there are some trash email services but the config value needs to force it everywhere or nowhere.

Silic0nS0ldier:
Quite a few services allow multiple secondary emails to be added to a account for security alerts (e.g. security alerts and account recovery). Might be worth keeping that in mind if database changes are required (and they probably will be).

amosfolz:
How about if we moved towards using a polymorphic many-to-many relationship to allow for more than one email per user? The idea would be to allow for easier user customization or inside UserFrosting in the future. (E.g. could be used for phone numbers, addresses, etc...)

lcharette:
That might introduce complexity since the email can be used for login too. A relation for any stuff that need verification could work though. Or if you want to keep old email records.

PR #1113, opened 2020-10-09
Security Fix for E-mail Verification Bypass - huntr.dev 

huntr-helper:
https://huntr.dev/users/alromh87 has fixed the E-mail Verification Bypass vulnerability 🔨. alromh87 has been awarded $25 for fixing the vulnerability through the huntr bug bounty program 💵. Think you could fix a vulnerability like this?
           
Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/UserFrosting/pull/3
Vulnerability README | https://github.com/418sec/huntr/blob/staging/bounties/packagist/userfrosting/userfrosting/3/vulnerability.json

### ⚙️ Description *

UserFrosting is following strict email verification but can be bypassed by changing email in account settings.

Bypass is fixed by requiring verification before updating to new email.

### 💻 Technical Description *

Existent verification mecanism was extended to handle email address change. If mail-verification is enabled in settings, acount settings update will store newEmail as a request and send an email with link for verification, after link is open by new email owner settings will be updated to new email.

### 🐛 Proof of Concept (PoC) *

1. Setup UserFrosting repo
2. Create a user using valid email
3. Verify new acount by following verification link
4. Log in and go to My Account
5 Change the email and save
6. New email will be enabled without verification
7. Now you can login with unverified email 

userfrostingEmailPOC0

userfrostingEmailPOC1

### 🔥 Proof of Fix (PoF) *

After fix verification email is sent to requested address and account information won't be updated untill verification is completed:

NewEmailSent

To verify follow received link:
NewEmailLink

After verification new email can be used, account information is updated:
NewEmailOK

### 👍 User Acceptance Testing (UAT)

e-mail can be updated using verification and application works normally

### DIFF
--- a/app/sprinkles/account/config/default.php
+++ b/app/sprinkles/account/config/default.php
@@ -162,7 +162,18 @@ return [
         'check_username_request' => null,
         'password_reset_request' => null,
         'registration_attempt'   => null,
-        'sign_in_attempt'        => null,
+        'sign_in_attempt'        => [
+            'method'   => 'ip',
+            'interval' => 3600,
+            'delays'   => [
+                4 => 5,
+                5 => 10,
+                6 => 20,
+                7 => 40,
+                8 => 80,
+                9 => 600,
+            ],
+        ],
         'verification_request'   => null,
     ],
 
--- a/app/sprinkles/account/locale/en_US/messages.php
+++ b/app/sprinkles/account/locale/en_US/messages.php
@@ -39,6 +39,7 @@ return [
             '@TRANSLATION'  => 'Account settings',
             'DESCRIPTION'   => 'Update your account settings, including email, name, and password.',
             'UPDATED'       => 'Account settings updated',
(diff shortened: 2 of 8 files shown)


alexweissman:
Thanks for finding this and submitting a fix! My main concern is with adding a new column to the `users` table. Having the new email address there seems like poor encapsulation and generally a bad code smell. Would it not make more sense to add an `email` column to the `verifications` table?

We could even consider expanding this to allow users to have multiple email addresses, each of which may or may not be verified, and one email address designated as the primary address. Then we change the concept from "verified user" to "verified email address". Privileges (including the ability to sign in) could still be managed depending on whether the user's primary email address has been verified or not. See for example the [Github email settings page](https://github.com/settings/emails).

alexweissman (changes requested):
See comment.

Silic0nS0ldier:
Related: https://github.com/userfrosting/UserFrosting/issues/820

PR #1114, opened 2020-10-15
Deep extend when adding global query params in ufTable

alexweissman:
At the moment if you were to use `addParams` to set a global `filter` or `sort` for a table, it would overwrite all other filters and sorts applied to the table. It doesn't seem like this was a deliberate design choice when `addParams` was introduced in https://github.com/userfrosting/UserFrosting/commit/3159224e942d741b89c40cc8c52108e7e0cf504f. By using a deep extend instead, we can set global filters without overriding any sorts/filters set by the user. It does technically change the behavior, but I don't see a valid use case for the way it currently behaves.

### DIFF
--- a/app/sprinkles/core/assets/userfrosting/js/uf-table.js
+++ b/app/sprinkles/core/assets/userfrosting/js/uf-table.js
@@ -437,7 +437,7 @@
         $.extend(table.config.pager.ajaxObject.data, tableState);
 
         // Merge in any additional parameters
-        $.extend(table.config.pager.ajaxObject.data, this.settings.addParams);
+        $.extend(true, table.config.pager.ajaxObject.data, this.settings.addParams);
 
         return url;
     };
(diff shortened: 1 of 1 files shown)


Silic0nS0ldier:
Code looks fine, but I can't approve this. Merging options can be a deceptively complex process, as certain combinations of merged configurations can end up being invalid (which ends up requiring people passing configurations in having to explicitly specify more configuration options).

To properly review this I've need to first figure out the shape of the configuration option and what those options are.

alexweissman:
@Silic0nS0ldier In general agree, but in this case the parameters are only merged into a very specific portion of the config options. Namely, the `ajaxObject.data` key which is empty by default. See https://mottie.github.io/tablesorter/docs/example-pager-ajax.html and https://api.jquery.com/jQuery.ajax/#jQuery-ajax-settings. So, it shouldn't be possible to touch any of the other Tablesorter or `ufTable` parameters through this setting.

Silic0nS0ldier (approved):
Alright. LGTM. Just need to make sure the changelog mentions what has changed (to address the remote change someone is impacted).

lcharette:
Since the base got messed up, I merged it manually in Hotfix: 3310530db83229f37bf45963fc8e0e471c7d881d

PR #1118, opened 2020-11-14
4.4.4 Release

lcharette:


Silic0nS0ldier (approved):

Silic0nS0ldier:
Source looks good, but Travis seems to have caught something. Looks to be a flaky test given the PR build event succeeded.
https://travis-ci.org/github/userfrosting/UserFrosting/jobs/743784573

lcharette:
Looks like a false negative from a duplicate factory value, restarted the test to make sure.

Silic0nS0ldier:
@lcharette As in a factory created a randomised user that ended up having something in common with another generated user, resulting in a fail state? If that's the case, a-okay with me. Probably worth flagging for a follow-up though, randomisation is good for flushing out bugs but is a poor fit for unit tests IMO.

Silic0nS0ldier:
@lcharette As in a factory created a randomised user that ended up having something in common with another generated user, resulting in a fail state? If that's the case, a-okay with me. Probably worth flagging for a follow-up though

lcharette:
Exactly. It's a limitation in Factory Muffin afaik, which has a limited set of unique elements. But could probably also be traced back to a poor test design where a DB is reused when it shouldn't (or not reused enough).

Issue #1117, opened 2020-11-14
Incompatible with Composer 2


lcharette:
```
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - wikimedia/composer-merge-plugin dev-master requires composer-plugin-api ^1.1 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - wikimedia/composer-merge-plugin[v1.4.0, ..., v1.4.1] require composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - wikimedia/composer-merge-plugin 1.5.x-dev is an alias of wikimedia/composer-merge-plugin dev-master and thus requires it to be installed too.
    - Root composer.json requires wikimedia/composer-merge-plugin ^1.4.0 -> satisfiable by wikimedia/composer-merge-plugin[v1.4.0, v1.4.1, 1.5.x-dev (alias of dev-master)].

You are using Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report a plugin-issue to ask them to support Composer 2.
```

This is dependent of an issue with [wikimedia/composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin), and can only be solved once https://github.com/wikimedia/composer-merge-plugin/pull/189#issuecomment-719514950 is merged and released. 

Other reference : https://github.com/wikimedia/composer-merge-plugin/issues/184

lcharette:
We should be able to fix this now by upgrading `wikimedia/composer-merge-plugin`  to this new release : 
https://github.com/wikimedia/composer-merge-plugin/releases/tag/v2.0.1

Silic0nS0ldier:
_And_ raising our minimum PHP version from 7.1 to 7.2 (as per plugin requirements). Otherwise looks to a fairly straightforward update thankfully.

lcharette:
We're still supporting 7.1? Might as well release 4.5 while at it. Do you have any other change / BC that requires a major bump ?

Silic0nS0ldier:
Yeah, NodeJS for ESM support. Not ready for prime time yet, but raising will let it release more easily.
Min is `^12.17.0 || >=13.2.0`.

lcharette:
> Yeah, NodeJS for ESM support. Not ready for prime time yet, but raising will let it release more easily.
> Min is `^12.17.0 || >=13.2.0`.

[...]

PR #1120, opened 2020-11-20
Composer2 fix

lcharette:
Related to #1117. Testing fix from https://github.com/wikimedia/composer-merge-plugin/pull/189. Should probably stay in draft until official wikimedia/composer-merge-plugin is released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ before_install:
 
 before_script:
   # install deps and UF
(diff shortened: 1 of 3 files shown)


lcharette:
image Travis ans it's limitations when comes to using VCS repo with composer... I did test manually on the VM and the temp fix works for now. I'll still wait for the official release.

Issue #1098, opened 2020-05-15
Use Laravel's `env` helper instead of `getenv`


alexweissman:
The solution here is to not use the native `getenv` at all, but rather Laravel's [`env` helper](https://laravel.com/docs/7.x/helpers#method-env) which takes a default value as its second parameter. So this can just be `env('CSRF_ENABLED', true)`. This also has the nice side effect of allowing falsy values like `false` in the environment variable.

_Originally posted by @alexweissman in https://github.com/userfrosting/UserFrosting/pull/976/review_comment/create_

alexweissman:
Labeling this as a bug because `CSRF_ENABLED=false` does not work with our current implementation using `getenv`.

lcharette:
Is it compatible with DotEnv ?

lcharette:
Note for later : https://github.com/illuminate/support/blob/master/Env.php

alexweissman:
Yes, in fact Laravel uses DotEnv as well.  `env` is just a wrapper around the `getenv` function which DotEnv overloads.

lcharette:
And what's the issue with `CSRF_ENABLED` ? How can we reproduce ?

alexweissman:
It's not really specific to `CSRF_ENABLED`, it's just that that variable happens to be one that takes a true/false value. You can try creating any new variable in `.env` and setting it to `false`. If you import it with `getenv`, you will get the string `"false"` (which evaluates to a boolean true) rather than an actual boolean value of `false`.

PR #1121, opened 2020-11-24
Replace getenv() with env()

alexweissman:
See https://github.com/userfrosting/UserFrosting/issues/1098. Not sure if we want to merge this in for a revision or save for another minor version.

### DIFF
--- a/app/sprinkles/core/src/Bakery/DebugCommand.php
+++ b/app/sprinkles/core/src/Bakery/DebugCommand.php
@@ -49,10 +49,10 @@ class DebugCommand extends BaseCommand
 
         // Perform tasks & display info
         $this->io->definitionList(
-            ['UserFrosing version'  => \UserFrosting\VERSION],
+            ['UserFrosting version'  => \UserFrosting\VERSION],
             ['OS Name'              => php_uname('s')],
             ['Project Root'         => \UserFrosting\ROOT_DIR],
-            ['Environment mode'     => getenv('UF_MODE')],
+            ['Environment mode'     => env('UF_MODE', 'default')],
             ['PHP Version'          => $this->checkPhpVersion()],
             ['Node Version'         => $this->checkNodeVersion()],
             ['NPM Version'          => $this->checkNpmVersion()]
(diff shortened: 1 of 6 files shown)


app/sprinkles/core/src/Bakery/DebugCommand.php
```diff
@@ -49,10 +49,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         // Perform tasks & display info
         $this->io->definitionList(
-            ['UserFrosing version'  => \UserFrosting\VERSION],
```
lcharette, line 52: https://i.kym-cdn.com/entries/icons/mobile/000/030/710/dd0.jpg

lcharette:
So `env` is a magic helper method of Laravel? Not sure I like that kind of magic, after the [`str_contain`](https://github.com/laravel/framework/issues/33045) fiasco. I wonder if they have a namespaces equivalent, like `Str::` method ?

If it doesn't change `.env` value, I don't see why we couldn't merge in hotfix ?

lcharette (approved):

lcharette (changes requested):
Actually, there is a namespaced equivalent : https://github.com/illuminate/support/blob/master/Env.php 

I think it would be better long term...

alexweissman:
I would say we should just follow the same pattern as Laravel, which uses the `env` helper in config files and tests: https://github.com/laravel/laravel/blob/8.x/config/database.php. I think this makes sense because the namespace could change in subsequent versions of Laravel but the helper method would still be independent of the namespace.

They do use `Env::` in namespaced code, if we wanted to be consistent there. Note that unlike the `str_` helpers, `env` has not been deprecated.

lcharette (approved):
Fair enough

lcharette:
Note there was actually an issue with this PR that appeared after merge. See temp fix in https://github.com/userfrosting/UserFrosting/commit/926b9be79971a0d3c6d7b8765d4dc7dcb15f55dd

Error : https://github.com/userfrosting/UserFrosting/actions/runs/628080953

Looks like env doesn't use the loaded env right above : 
https://github.com/userfrosting/UserFrosting/blob/926b9be79971a0d3c6d7b8765d4dc7dcb15f55dd/app/sprinkles/core/src/ServicesProvider/ServicesProvider.php#L235-L245

PR #1122, opened 2020-11-27
Implement class table inheritance

alexweissman:
See https://martinfowler.com/eaaCatalog/classTableInheritance.html. The new (optional) trait HasAuxModel and scope AuxModelScope greatly simplify the process for extending the User model (see https://learn.userfrosting.com/recipes/extending-the-user-model). Simply define the aux migration/model, and include `HasAuxModel` and set the `auxType` property in your extended user model.

Note that this does not affect the current functionality in any way, as it is not included in any of the default UF models. It is however a major feature that would change the "extending the user model" recipe, so we could wait until the next minor version.

### DIFF
--- a/app/sprinkles/core/src/Database/Concerns/HasAuxModel.php
+++ b/app/sprinkles/core/src/Database/Concerns/HasAuxModel.php
@@ -114,8 +114,8 @@ trait HasAuxModel
     /**
      * Set a given attribute on the main model or aux model.
      *
-     * @param  string $key
-     * @param  mixed  $value
+     * @param string $key
+     * @param mixed  $value
      *
      * @return mixed
      */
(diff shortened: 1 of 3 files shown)


lcharette (approved):
Looks good to me. Definitively useful.

Only bummer is the main User model still need to be extended by the sprinkle. Could also be beneficial if we could have multiple aux linked model. Then we could have them registered in the config or serviceProviders. But much of that could be added in later revision.

As far as which branch to release this, I guess it could be either hotfix or develop. Documenting the feature might be the deciding factor when we release it.

alexweissman:
Yep, I'm already using it in my projects to model different types of users with different aux tables. No matter what you'll have to extend a base model and use the class mapper in some places to override the base `User` model, for example. I really wish PHP8 would have introduced dynamic identifiers as a feature but 🤷

PR #1124, opened 2020-11-29
Implement findInt

alexweissman:
Convenience method for validating and retrieving a single model by integer primary key. Notice that this also solves the issue where MySQL automatically converts strings to the integer 0, which can cause misleading error messages if you don't validate first: https://stackoverflow.com/questions/21762075/mysql-automatically-cast-convert-a-string-to-a-number


lcharette (approved):

lcharette:
Looks good. Should probably be merged in develop for 4.5 in case someone extend the base sprunje already ?

alexweissman:
Yeah that makes sense.

lcharette (approved):

lcharette:
N.B.: Had to switch target to master and back to develop as branch wasn't in sync for some reason. The filling Node 15 test is currently a known issue.

PR #1126, opened 2020-11-29
Force Composer 1.0 for Docker

ssnukala:
Some of the UF Dependencies are not yet compatible with Composer 2.x so forcing downgrade to Composer 1.x


lcharette:
Thanks. I'll merge in hotfix.

PR #1128, opened 2020-12-03
Update error.html.twig - add container

SarahBaghdadi:
The error page content should be in a container so that it doesn't live right up against the edge of the DOM. Adding the `container` class to this div will fix it.


lcharette (approved):
I'll trust your work on this

Silic0nS0ldier:
Before: 
Screen Shot 2020-12-05 at 1 22 18 pm

After: 
Screen Shot 2020-12-05 at 1 23 08 pm

Only visual difference I noticed is page content having a little more padding. Not quite clear on the purpose of the change, but CSS is largely a black box to me so happy to defer to those more experienced.

LGTM.

Silic0nS0ldier (approved):

PR #1132, opened 2020-12-25
Replace Travis with GitHub Actions

lcharette:
There's a failure in the tests because there's actually an issue with the Windows support. Should be squashed merged if possible.


Silic0nS0ldier (approved):
The degree of customisation needed for each platform is disappointing (but not unexpected).

With the note that we'll need to revisit Windows support, this LGTM.

lcharette:
> The degree of customisation needed for each platform is disappointing (but not unexpected).

N.B.: The number of jobs could be lower and some merged together, but it requires some "if" and it's easier this way.

PR #1138, opened 2021-03-04
Raise minimum node version

Silic0nS0ldier:


Silic0nS0ldier:
CI is on fire, but the bits relevant to this are passing. With the one exception being NodeJS 15 on Windows (hit the same issue under different circumstances recently, I suspect something changed with v15) which I will follow up later as part of ESM-ification of the assets pipeline (see https://github.com/orgs/userfrosting/projects/2).

Going to get this merged now to try and keep things moving forward.

Also as an aside (so that its written _somewhere_), we probably don't need to trigger the CI on PR given its already done on push. This is something I'm doing for the JS repos, and the codecov integrations are working fine. https://github.com/userfrosting/gulp-uf-bundle-assets/pull/305

lcharette:
Ci will probably be fixed once I merge in hotfix...

> Also as an aside (so that its written _somewhere_), we probably don't need to trigger the CI on PR given its already done on push. This is something I'm doing for the JS repos, and the codecov integrations are working fine. [userfrosting/gulp-uf-bundle-assets#305](https://github.com/userfrosting/gulp-uf-bundle-assets/pull/305)

Won't the PR CI be useful for a PR coming from a Fork not having CI enabled ?

Issue #1140, opened 2021-03-09
Asset dependency install fails on Windows with NPM v7


lcharette:
See here : https://github.com/userfrosting/UserFrosting/runs/2061948433?check_suite_focus=true#step:7:47

```
Installing frontend vendor assets
---------------------------------

> npm run uf-assets-install

> uf-assets-install
> ./node_modules/.bin/gulp assetsInstall

'.' is not recognized as an internal or external command,
operable program or batch file.

Error: ] assets installation has failed       
                                                                         
Error: Process completed with exit code 1.
```

Should be the only missing piece for 4.5.0 release (beside usual stuff).

Silic0nS0ldier:
I forgot about that. Fix should be to remove `./` from `build/package.json` scripts should do the trick. NPM v7 would be responsible for this.

Silic0nS0ldier:
With this being a bug on develop and having been resolved there, I'm closing this.

PR #1142, opened 2021-03-10
Improved PHP, Node and NPM version check

### LINKED ISSUE #1141: Missing node version in Bakery Debug command
I noticed the Bakery debug command doesn't show the current node version anymore : 

```
---------------------- --------------------------------------------- 
  UserFrosting version   4.5.0-dev                                    
  OS Name                Linux                                        
  Project Root           /home/runner/work/UserFrosting/UserFrosting  
  Environment mode       debug                                        
  PHP Version            7.2.34-18+ubuntu20.04.1+deb.sury.org+1       
  Node Version                                                        
  NPM Version            6.14.11                                      
 ---------------------- --------------------------------------------- 
```

For example : 
https://github.com/userfrosting/UserFrosting/runs/2061948500?check_suite_focus=true#step:12:24

lcharette:
Fix #1141


app/sprinkles/core/src/Exceptions/VersionCompareException.php
```diff
@@ -0,0 +1,52 @@
...
+ * @copyright Copyright (c) 2019 Alexander Weissman
+ * @license   https://github.com/userfrosting/UserFrosting/blob/master/LICENSE.md (MIT License)
+ */
+
+namespace UserFrosting\Sprinkle\Core\Exceptions;
+
+use LogicException;
+
+/**
+ * Version Compare Exception. Used when a class name is dynamically invoked, but the class does not exist.
```
Silic0nS0ldier, line 16: This description makes no sense.

app/sprinkles/core/src/Util/CheckEnvironment.php
```diff
@@ -351,36 +366,39 @@ public function checkPermissions()
...
-        // Check PHP version
-        if (version_compare(phpversion(), \UserFrosting\PHP_MIN_VERSION, '<')) {
-            $problemsFound = true;
+        try {
+            $this->checkPhpVersion();
+        } catch (VersionCompareException $e) {
             $this->resultsFailed['phpVersion'] = [
                 'title'   => "<i class='fas fa-code fa-fw'></i> You need to upgrade your PHP installation.",
-                'message' => "I'm sorry, UserFrosting requires version " . \UserFrosting\PHP_MIN_VERSION . ' or greater.  Please upgrade your version of PHP, or contact your web hosting service and ask them to upgrade it for you.',
+                'message' => "I'm sorry, UserFrosting requires version " . $e->getContraint() . ' or greater. Version ' . $e->getVersion() . ' detected. Please upgrade your version of PHP, or contact your web hosting service and ask them to upgrade it for you.',
```
Silic0nS0ldier, line 379: Given we are using a semver range, the message should probably be revised. "requires version ^7.2 or greater" doesn't make much sense.

app/sprinkles/core/src/Util/CheckEnvironment.php
```diff
@@ -351,36 +366,39 @@ public function checkPermissions()
...
-            ];
+
+            return true;
         }
 
-        return $problemsFound;
+        // No problem found !
+        $this->resultsSuccess['phpVersion'] = [
+            'title'   => "<i class='fas fa-code fa-fw'></i> PHP version checks out!",
+            'message' => "You're using PHP " . \UserFrosting\PHP_MIN_VERSION . 'or higher.  Great!',
```
Silic0nS0ldier, line 389: As before, message should reflect constant being a semver range expression.

app/sprinkles/core/src/Util/VersionCheck.php
```diff
@@ -0,0 +1,105 @@
...
+     * @throws VersionCompareException If contraint version is not matched.
+     *
+     * @return string The current PHP Version
+     */
+    protected function checkPhpVersion(): string
+    {
+        $phpVersion = (string) phpversion();
+
+        if (!Semver::satisfies($phpVersion, \UserFrosting\PHP_MIN_VERSION)) {
+            $exception = new VersionCompareException('UserFrosting requires php version ' . \UserFrosting\PHP_MIN_VERSION . ' or above. You have ' . $phpVersion . ". You'll need to update you PHP version before you can continue.");
```
Silic0nS0ldier, line 33: Semver expresses a range.

app/sprinkles/core/src/Util/VersionCheck.php
```diff
@@ -0,0 +1,105 @@
...
+     * @throws VersionCompareException If contraint version is not matched.
+     *
+     * @return string The current PHP Version
+     */
+    protected function checkPhpDeprecatedVersion(): string
+    {
+        $phpVersion = (string) phpversion();
+
+        if (!Semver::satisfies($phpVersion, \UserFrosting\PHP_RECOMMENDED_VERSION)) {
+            $exception = new VersionCompareException('While your PHP version is still supported by UserFrosting, we recommend version ' . \UserFrosting\PHP_RECOMMENDED_VERSION . ' or above as ' . $phpVersion . ' will soon be unsupported. See http://php.net/supported-versions.php for more info.');
```
Silic0nS0ldier, line 55: semver expresses a range

Silic0nS0ldier (changes requested):
Mainly exception and output phrasing corrections reflecting semver ranges being self descriptive regarding what they cover. NPM and Node messages provide an example.

lcharette:
(Well, that's an interesting failure)

app/sprinkles/core/src/Exceptions/VersionCompareException.php
lcharette: Forgot to change after copy and paste...

Silic0nS0ldier (approved):
LGTM

lcharette:
Fixed the exception messages... I'll have to think of something for the PHP compare failure... Fun fact, technically it should fail on `version_compared` to... 

(glad I decided to do a proper PR for this one...)

lcharette:
Hum... I could probably add proper testing for the Trait too while at it...

lcharette:
@Silic0nS0ldier Does it make sense that `check*Version` methods returns the version as string? Throwing an exception when the check fails has the advantages of passing more info (message, target and constraint), but having a string if it's fine? Would returning `true` be better, having another method that does return the raw version instead? It would mean the bakery command and whatnot would a bit more complex...

https://github.com/userfrosting/UserFrosting/pull/1142/files#diff-29dc1db62db9ce87a33a02d791f8e8ac2a731dbe815952209bf80c20b3f494e7R28-R41

Silic0nS0ldier:
Returning a boolean vs. throwing an exception is generally better for perf. But as for code complexity, perhaps a rename is in order. Something like `getValid*Version`.

Semantically though, `check*Version` doesn't make much sense returning something other than a boolean or throwing. (and `assert*Version` would be better imo).

lcharette:
I've rework the whole thing. Moved the Trait to a static class/methods. Let me know what you think. ç

Meanwhile I'm still looking into the `7.2.34-18+ubuntu20.04.1+deb.sury.org+1` issue

lcharette:
> Meanwhile I'm still looking into the `7.2.34-18+ubuntu20.04.1+deb.sury.org+1` issue

Might be an issue to be solve in Semver. I opened an issue : https://github.com/composer/semver/issues/125
We could implement our own sanitizer, but I believe it would be easier/better to fix it on Semver side.

lcharette:
Ok, I'm finally happy with this. Mocking the built-in function in the tests makes much more sense. As pointed in semver issue, the php version is technically invalid, so I get it makes sense to sanitized on our side. And I prefer doing in in `getPhpVersion`.

app/sprinkles/core/src/Util/CheckEnvironment.php
```diff
@@ -351,36 +364,39 @@ public function checkPermissions()
...
-        $problemsFound = false;
-
-        // Check PHP version
-        if (version_compare(phpversion(), \UserFrosting\PHP_MIN_VERSION, '<')) {
-            $problemsFound = true;
+        try {
+            VersionValidator::validatePhpVersion();
+        } catch (VersionCompareException $e) {
             $this->resultsFailed['phpVersion'] = [
                 'title'   => "<i class='fas fa-code fa-fw'></i> You need to upgrade your PHP installation.",
```
Silic0nS0ldier, line 376: We limit the upper bound as well (PHP 8 would register as an incompatible version). Message should be updated to reflect this.

Silic0nS0ldier (changes requested):
Just the one nit-pick. Otherwise LGTM.

app/sprinkles/core/src/Util/CheckEnvironment.php
lcharette: Good point

lcharette: Updated the title to reflect the failure one

Silic0nS0ldier (approved):

Silic0nS0ldier:
Codecov is angry, but LGTM.

lcharette:
Yeah, I added a bit more coverage by including the exception, but the rest of patch diff hit is because of the Bakery commands & CheckEnvrionnement have added lines, but they are not covered because the Bakery command doesn't have existing tests...

(For future reference, the other CI failure is linked to #1140)

PR #1156, opened 2021-04-24
Translate alert stream messages at read time

### LINKED ISSUE #811: WELCOME message not in user's locale
The first page (dashboard) is rendered in the user's locale but the alert sent along with it is rendered in the default locale.

Forum discussion:
https://forums.userfrosting.com/t/welcome-message-not-in-users-locale/93/4

Silic0nS0ldier:
This revises the alert stream logic to delay actual translation of messages until they are read.

Example with English browser locale, and Chinese account locale.
image

This is marked as a prototype for now as I'm not 100% entirely settled on the data shape.

### DIFF
--- a/app/sprinkles/core/src/Alert/AlertStream.php
+++ b/app/sprinkles/core/src/Alert/AlertStream.php
@@ -97,9 +97,15 @@ abstract class AlertStream
             throw new \RuntimeException('No translator has been set!  Please call MessageStream::setTranslator first.');
         }
 
-        $message = $this->translator->translate($messageId, $placeholders);
+        $messages = $this->messages();
+        $messages[] = [
+            'type'    => $type,
+            'message_id' => $messageId,
+            'placeholders' => $placeholders,
+        ];
+        $this->saveMessages($messages);
 
-        return $this->addMessage($type, $message);
+        return $this;
     }
 
     /**
@@ -111,10 +117,31 @@ abstract class AlertStream
      */
     public function getAndClearMessages()
     {
-        $messages = $this->messages();
+        $messages = $this->translateMessages($this->messages());
         $this->resetMessageStream();
 
         return $messages;
+    }
+
+    /**
+     * Translates messages that have a message id instead of a message.
+     */
+    protected function translateMessages(array $messages)
+    {
+        $translatedMessages = [];
+
+        foreach ($messages as $message) {
+            if (array_key_exists('message_id', $message)) {
+                $translatedMessages[] = [
+                    'type' => $message['type'],
+                    'message' => $this->translator->translate($message['message_id'], $message['placeholders']),
+                ];
+            } else {
+                $translatedMessages[] = $message;
+            }
+        }
+
+        return $translatedMessages;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lcharette:
I agree with the process of translating when messages are displayed. Problem is now, `addMessageTranslated` and `addMessage` do 95% the same thing. And `addMessage` will now return a translated message (change in behavior). 

Actually, that whole class requires a rewrite:
-  We need an interface, not only an abstract class. 
- `messages` method needs to be renamed the proper `getMessages`. 
- A message should be an object (with appropriate interface). Translator could then be done in the message object. Type could be defined as part of the object (eg. `errorMessage implement Message`, `sucessMessage implement Message`). PRS-14 events could be a good inspiration for this. 

I was planning on moving `Alert` to the framework repo as of UF 5. We could do the rewrite in 5.0.

Silic0nS0ldier:
Write is probably best reserved for v5. Ideally whatever solution we land on would avoid requiring modifications to alert stream implementations when a new alert type is introduced.

> And `addMessage` will now return a translated message (change in behavior).

No it doesn't? While `addMessage` and `addMessageTranslated` do have very similar source now, their return type remains `$this`. The only API surface change here is the addition of the protected function `translateMessages`.

lcharette:
> No it doesn't? While `addMessage` and `addMessageTranslated` do have very similar source now, their return type remains `$this`. The only API surface change here is the addition of the protected function `translateMessages`.

Api doesn't change, but behavior does. If you _don't_ want a translated message, it's changed.

Silic0nS0ldier:
> If you _don't_ want a translated message, it's changed.

`addMessage(...)` has no behavior changes. While messages added by this will be checked in `translateMessages(...)`, they will not be modified as they are already in the desired shape (`['type' => string, 'message' => string]`).

`addMessageTranslated(...)` and `messages(...)` (and by extension `getAndClearMessages(...)`) is where the behaviour change lies, that being translation occuring at read time as opposed to write.

---

On a different note, some reminders for myself;

1. Messages are being processed by `translateMessages(...)` twice. Not much can be done about the API surface right now, so the solution is to remove the call from `getAndClearMessages(...)` such that implementations are required to call `translateMessages(...)` on the retrieved messages. In UF5, we can revise the API surface here to something that violates less code conventions.
2. Audit usages of `addMessageTranslated(...)`, ensure only the bare minimum placeholder values are passed in since they will now all be persisted for a time (in the case of the welcome message, the entire exportable user data is included).
3. Move the translator check into the `translateMessages(...)` function from `addMessageTranslated(...)`.

Silic0nS0ldier:
Fun fact, a `UserMessage` class already exists in the support repo specifically for things like the alert stream. Just noting here in case it helps future planning.

lcharette:
> Fun fact, a `UserMessage` class already exists in the support repo specifically for things like the alert stream. Just noting here in case it helps future planning.

Good catch. And it's currently used for [Exceptions](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/admin/src/Controller/RoleController.php#L360). The [`UserMessage`](https://github.com/userfrosting/support/blob/master/src/Message/UserMessage.php) class does required a modernization (public property 👎) and interface.

lcharette:
I manually integrated part of this in [dd98b7f](https://github.com/userfrosting/framework/commit/dd98b7f6259fa020ce602e6b9b34a1774477798e) for the 5.1 release.

PR #1155, opened 2021-04-24
Misc. improvements to assets pipeline

Silic0nS0ldier:
* Type validation of build scripts
* Removed redundant properties from in-box bundle definitions
* Bumped `jquery-validation` and `urijs` up to bring in vulnerability fixes
* Split gulp logger adapter into its own package
* Improved handling of multi-line log items
* Native ESM (asset processing starts much more quickly, this extends to many direct dependencies)
* Typo swatting
* Updating placeholder image URLs (old service now redirects, placeholders failed to render)
* Updates NodeJS versions in CI to use newest version in major version range
  All versions save for the newest in a given supported major NodeJS release is effectively end of life. While we aim for compatibility across an entire major version line, it is impractical to enforce this in CI since it means running ancient versions which may be full of vulnerabilities, bugs and subtle behaviour changes (e.g. CJS module imports having named import support introduced)..

This is mostly complete, remaining work includes;
- [x] Release 1.0.0 of `gulp-prune`
- [x] Release 1.0.0 of `@userfrosting/ts-log-adapter-gulplog`

PR #941, opened 2019-02-20
ufModel architecture update

Silic0nS0ldier:
This was reverted quite a long time ago due to bugs with the plan being to review this again later on. Leaving this as a stale undocumented branch isn't going to lead to this getting done, so I've thrown this draft PR together.

---

This has been refactored (to reduce dependence on `this` and provide better documentation of types) and lightly tested. Some redundant code paths have also been eliminated.

As an added bonus, less dependence on `this` has resulted in a significant size reduction post-compilation.

PR #1160, opened 2021-05-01
Release 4.6.0

### LINKED ISSUE #740: Migrate ufModal to jQuery Boilerplate
We tried this [before](https://github.com/userfrosting/UserFrosting/issues/646), but there was a [bug](https://github.com/userfrosting/UserFrosting/issues/732) that forced us to revert to the old plugin structure.

### LINKED ISSUE #1074: Update to the latest Laravel (8.x)
I realize this was done in #907 however #907 updated Illuminate to 5.8.x (current at the time). Laravel recently released 7.x

Whats the time frame for this project on updating libraries like Laravel (Illuminate) to newer versions? 

Would we be able to look into going to Illuminate 7.x?
- https://laravel.com/docs/6.x/upgrade
- https://laravel.com/docs/7.x/upgrade

As for something like Slim (3 to 4) that might be a way more drastic upgrade than Illuminate 5.8 to 7 (See Slim 3 => 4: http://www.slimframework.com/docs/v4/start/upgrade.html)

### LINKED ISSUE #1130: PHP 8 Support
UF's support of PHP 8 is unknown at this time as we do not have it covered by the existing test matrix. We'll likely need to update some dependencies in order to support PHP 8, and may also have to raise the minimum version.

### LINKED ISSUE #1131: Theming, currentUser, and the assets service (deprecate per user theme)
Our current implementation of theming allows the `currentUser` service to dynamically load a sprinkle based on the `users.theme` column in the database. Since theme sprinkles can potentially contain assets, the `assets` service is overridden to force the `currentUser` service to boot up on every request.

This created a potential performance issue in that the database had to be queried for every request - including requests for raw assets. To solve this, the current user object is now automatically cached on disk (or in redis) for a period of time (https://github.com/userfrosting/UserFrosting/commit/522f99e57a0980bc73aab4d836adb480f29da8b7). 

This caching is done in the `Authenticator:: validateUserAccount` method. However this does not account for any custom implementations outside of the `Authenticator` class. For example, I override the `currentUser` service to dynamically load a user subtype model depending on the value of a user type column. This causes the database to be hit via the `currentUser` service, which is forced to boot by the `assets` service.

I actually don't even use the theming feature, so one possibility would be to simply disable theming somehow so that `assets` doesn't need to boot `currentUser`. Or, I could have the ability to define my own custom `Authenticator` class that caches my subtype model along with the original user model. This is not possible at the moment because there are a few hard dependencies on the `Authenticator` class instead of depending on an interface.

My final option is to check the request path in my `currentUser` model and skip loading the subtype model if the request is for a raw asset. This seems rather kludgy though.

Does it really make sense to implement per-user themes on the sprinkle level at all? Also, we should remove the hard dependencies on `Authenticator` and replace with an interface.

lcharette:
~[This looks like our fastest released yet]~ 

Address : 
- Fix #1074
- Fix #1130
- Fix #1131 
- Fix #1155
- Fix #941 
- Fix #740

Steps to complete before merging : 
- [x] Update learn
- [x] Fix issue with Cache on Windows : https://github.com/userfrosting/framework/runs/2462341801?check_suite_focus=true#step:8:32
- [x] Fix code coverage on `userfrosting/framework`
- [x] Release `userfrosting/framework`
- [x] Update version in `app/defines.php`
- [x] Update from `hotfix` (#1165) 
- [x] Update `composer.json`
- [x] Merge #1155
- [x] Merge #941
- [x] Make sure CHANGELOG is up to date
- [x] Create 4.5 Branch 
- [x] Deploy Learn branch


Silic0nS0ldier (approved):

Silic0nS0ldier:
Amazing work @lcharette! So much for me getting some assets improvements out in the next minor version bump! 😅

amazing

Silic0nS0ldier (dismissed):

Silic0nS0ldier (dismissed):

Silic0nS0ldier (approved):

Silic0nS0ldier (approved):

Issue #1069, opened 2020-03-08
DatabaseTests only runs on in-memory database


lcharette:
https://github.com/userfrosting/UserFrosting/blob/9d3cd74432931696623ad9ce1f97600eb7574344/app/sprinkles/core/tests/Integration/Database/DatabaseTests.php#L23

Giving this is hardcoded, it doesn't use the `TEST_DB` env defined in [Travis config](https://github.com/userfrosting/UserFrosting/blob/9d3cd74432931696623ad9ce1f97600eb7574344/build/before_install.sh#L19). This means this test is only run with in-memory database.

This is not a simple fix, as it will requires : 

- [ ] A new method in `RefreshDatabase` Trait to create an empty database;
- [ ] Remove hardcoded string [throughout](https://github.com/userfrosting/UserFrosting/blob/9d3cd74432931696623ad9ce1f97600eb7574344/app/sprinkles/core/tests/Integration/Database/DatabaseTests.php#L1216)

Might need to apply same strategy to `DatabaseMigratorIntegrationTest`.

PR #1175, opened 2021-08-05
Several fixes for tests

ktecho:
At this point, all the tables from the standard migration have been created in previous tests, so tables like `users` or `roles` cannot be created again.

### DIFF
--- a/app/sprinkles/core/tests/Integration/Database/DatabaseTests.php
+++ b/app/sprinkles/core/tests/Integration/Database/DatabaseTests.php
@@ -34,6 +34,8 @@ class DatabaseTests extends TestCase
         // Boot database
         $this->ci->db;
 
+        $this->schema($this->schemaName)->dropAllTables();
+
         $this->createSchema();
     }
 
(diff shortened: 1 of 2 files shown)


Silic0nS0ldier:
To confirm, this PR fixes the database tests so that they can be rerun successfully. Correct?

ktecho:
@Silic0nS0ldier No, this fixes are needed to run the tests in a separate database/connection than the main one.

I cannot explain why it works here in github, but it's needed for the tests to pass in a separate MySQL.

I was talking about this with @lcharette yesterday... Maybe the tests are always running in an "in memory" database so we don't need to clear the database between tests?

ktecho:
@Silic0nS0ldier I want to update the `docker-compose` file to run another MySQL for tests, so you can do `php bakery test` whenever you want and run the tests in the devel environment without having to run another set of containers.

I have it running in our project, but this fixes are needed for the tests to pass.

lcharette (changes requested):
Fun fact, I encountered the same issue while working on 5.0 and before looking at this in details... The main thing is DatabaseTest is forced to run on in-memory database due to an hardcoded string (https://github.com/userfrosting/UserFrosting/issues/1069). 

Long story short, this whole mess is fixed in [5.0](https://github.com/userfrosting/sprinkle-core/blob/8042a983a53a0a0720a10aad3d6170a8ffba646a/tests/Integration/Database/DatabaseTests.php#L121), but I'm still willing to merge that in 4.x if requested change is applied.

lcharette:
Closing this, as it's not relevant with UF5 anymore

PR #1176, opened 2021-08-10
generateRandomToken improvement.

michaellrowley:
The MD5 call altogether has been switched out for the hexadecimal representation of 16 CSRNG-generated bytes (the same amount at MD5 produces).
See [this](https://huntr.dev/bounties/5cc92fbb-5a63-4693-aa01-775692fb137e/).

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,0 +1,22 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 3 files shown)


lcharette:
N.B.: This will need to be properly (manually) tested to make sure the existing token (if any) are not impacted and Captcha still properly working.

PR #1188, opened 2021-12-12
Release 4.6.3

lcharette:


lcharette:
It's ready to merge as 4.6.3, but... whouldn't that be a "soft" breaking change for existing install ? 🤔

Silic0nS0ldier:
> whouldn't that be a "soft" breaking change for existing install

Absolutely, and the break that will be seen on vulnerable sites is "everything stops working because links are broken". From a semver perspective I hate it, from a security perspective its the only way the to be confident the issue is sufficiently addressed. Bad assumptions about data trustworthiness is such a pain, might be worth doing an audit before the next big release (assuming any of us have the time).

If we had a simpler update process (using composer) I'd say do a major version bump and rely on existing vulnerability alert mechanisms in GitHub (and various third party services like the one that tipped us off) to inform users which in such a setup would work well.

lcharette:
Agree. I don't have enough to justify a new 4.x. version now, and one will probably not be done soon as I'm actively working on 5.0, but this can't wait. I'll leave a warning on the release.

PR #1192, opened 2022-01-28
Fix for AccountControllerTest when registration is disabled

archey347:
Fixed `AccountControllerTest` so that it doesn't fail when the registration is disabled in the main configuration.


lcharette:
Thanks !

Issue #1143, opened 2021-03-14
Cannot create another user with the same username of a deleted user


anesuc:
Hi all,

I just found a bug where if I remove a user from Admin (might do the same any other way, haven't tested), and I go to create a new user with the same username, it will fail saying to check error logs (checked both userFrost and php logs and there was nothing). This happened using the same username but a different email as well (so it's linked to the username). I've currently broken my myphpadmin to be able to check whats going on in the DB but I might make an effort later if this is a harder issue to debug.

User frost version: 4.4.5
PHP version 7.4.3

Thanks,
Anesu

lcharette:
That's because the deleted user is "soft deleted", so it still exist in the db, which is the intended purposed. However, I agree, the ability to permanently delete a user from the UI would be useful, plus a better error message.

anesuc:
Alright. Yeah that would be useful! So deleting users directly in the users table will not break anything right?

sampenguin:
I would add to this that hard deleting a user seems to be broken in UF 4.6.1 (calling User::delete(true)). The code flow is that It seems to try to soft delete a user first, but something internal is silently failing in that path... there is no error output I can find in any logs other than a 500 code in the access log for the api call.

Delaney:
Hey @lcharette is this issue still open?

lcharette:
User are still soft deleted. Open to suggestion how to handle this in the UI / CLI / whatever.

anesuc:
I think there should be an option for an Admin in the Admin Dashboard to be able to permanently delete a user record. The option can be something specific along with a warning message etc thats fine.

Delaney:
If emails or usernames from deleted accounts can be used again, my idea was to modify those fields and concatenate the `deleted_at` timestamp as a string to them, probably with a separator. Permanent deletion could break stuff.

lcharette:
A hard delete would mean all activities and other related data would also need to be deleted. Otherwise you'll get record without an associated user. In some cases it could break things. For example: "Project was created by {name} on {date}"... Delete the username, the project still exist, but not the Name (unless you hardcode the Name string, but that can bring other issue, if the name can change ;) )

@Delaney solution is another solution, but what if User1 gets deleted, User2 is created with the same email/username, and then you need to re-enable User1?

There's no "perfect" solution, and I think it comes back to each app how they want to handle it. But I agree it should be more obvious when a username is denied because of a deleted user.

Delaney:
[...]

PR #1193, opened 2022-02-07
Fix PHP7.4+ compatibility

avsdev-cw:
PHP 7.4 modified the method_exists behaviour so that it no longer works as required. Alternative is to use is_callable, however this always returns true when __call is present (which it is). Laravel kindly provides a method for checking if a macro exists, so use that instead.

This fixes the exception in #1143 but more work needs to be done for the UI to access soft-deleted items for hard deleting


avsdev-cw:
see: See: https://www.php.net/manual/en/function.method-exists.php#124462
and: https://www.php.net/manual/en/function.is-callable.php Notes section

PR #1201, opened 2022-02-27
Fix for FindUniqueTest when class mappings are changed

archey347:
`FindUniqueTest` fails in the accounts sprinkle when class mappings are changed for `user` and `group`.


lcharette:
Thanks !

lcharette:
Thanks !

PR #1202, opened 2022-03-02
Allowing no default group on register

avsdev-cw:
With this patch, setting the default group to "null" in the configuration means that newly registered users default to not being in a group. Previously this caused a "Registration is broken" error.

Incidentally, the translation for "ACCOUNT.REGISTRATION_BROKEN" does not exist either...

### DIFF
--- a/app/sprinkles/account/src/Account/Registration.php
+++ b/app/sprinkles/account/src/Account/Registration.php
@@ -90,17 +90,22 @@ class Registration
         // Validate the userdata
         $this->validate();
 
-        // Set default group
-        $defaultGroup = $this->ci->classMapper->getClassMapping('group')::where('slug', $this->defaultGroup)->first();
+        if ($this->defaultGroup) {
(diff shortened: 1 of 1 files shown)


lcharette:
Sounds good. I'm just curious about this line : 

https://github.com/userfrosting/UserFrosting/blob/80a3e1622ed06c918712045bf83f5732726f6584/app/sprinkles/account/src/Account/Registration.php#L93

The config would need to be explicitly `false`? Testing for false, null and/or `''` might be better? 


Note in long term, I'm looking into setting the default group in the db or something completely different.

avsdev-cw:
That already checks for false & null (null evaluates to false). the only case it doesn't check is '' which could be added?

avsdev-cw:
@lcharette Any movement on this?

Thanks

lcharette:
I merged it. This might change in V5 anyway.

PR #1211, opened 2022-09-29
Add Tanzanian Swahili Language

MegamindAme:
Add Tanzanian Swahili Language in the default sprinkle locales


lcharette:
This won't be merged yet, as built in locale will be changed in future version (they're difficult to maintain).

PR #1212, opened 2022-10-01
StrykeSlammerII v420 patch 1

StrykeSlammerII:
At one point I completely tore down my DB and tried rebuilding via migrations, with my own sprinkles in place. Somehow Bakery insisted on running the v420/AddingForeignKeys migration before most of the v400 migrations, so of course it errored and Bakery choked. I've added here the relevant dependencies, and Bakery was able to correctly run all my migrations with these changes.


lcharette:
I agree it make way more sense to add those dependencies.

PR #1217, opened 2023-04-11
Integrating upstream Whoops issue 752/PR 751

StrykeSlammerII:
This fixes Integration test WhoopsRendererTest::testRenderWhoopsPage() and related issues with the Whoops error handler.  Creating a new Run instance here, as I can't find an existing one in UF's WhoopsRenderer.
(See related upstream issue https://github.com/filp/whoops/issues/752 and associated PR https://github.com/filp/whoops/pull/751)
(See also UF_ConfigManager issue https://github.com/lcharette/UF_ConfigManager/issues/8#issuecomment-1481604990)

### DIFF
--- a/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php
+++ b/app/sprinkles/core/src/Error/Renderer/WhoopsRenderer.php
@@ -22,6 +22,7 @@ use Whoops\Exception\Formatter;
 use Whoops\Exception\Inspector;
 use Whoops\Handler\Handler;
 use Whoops\Handler\PlainTextHandler;
+use Whoops\Run;
 use Whoops\Util\Misc;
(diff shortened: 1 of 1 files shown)


StrykeSlammerII:
This fix feels like a hack to me--it works, but I don't know why a new object is required. If there's an existing Whoops\RunInterface, we should use that instead. 
I couldn't find one quickly, but I'm not familiar with UF's internals.

lcharette:
It is a hack, but I'm confortable with this fix. FYI, a permanent solution is already implemented in V5 (it use the `PrettyPageHandler` and `Run` directly) : 
https://github.com/userfrosting/sprinkle-core/blob/712bcc617a8fc98d3006539e0738969de8e7f36e/app/src/Error/Renderer/PrettyPageRenderer.php

PR #1222, opened 2023-04-19
Add utc param to dateFormat

phillmac:
Is this a sane feature we should have?
See 
https://github.com/tcort/handlebars-dateformat/issues/2
and 
https://github.com/tcort/handlebars-dateformat/commit/9d5021f639421e2b02db8c99b4e2981b0272dcbd
and
https://github.com/tcort/handlebars-dateformat/commit/9c3b4e9d41eb346b4c56ba7d84515124dc53d877

### DIFF
--- a/app/sprinkles/admin/assets/userfrosting/js/handlebars-helpers.js
+++ b/app/sprinkles/admin/assets/userfrosting/js/handlebars-helpers.js
@@ -60,10 +60,10 @@ Handlebars.registerHelper('calc', function (v1, operator, v2, options) {
  * usage: {{dateFormat creation_date format="MMMM YYYY"}}
  * @requires momentjs http://momentjs.com/
  */
-Handlebars.registerHelper('dateFormat', function(context, block) {
+Handlebars.registerHelper('dateFormat', function(context, block, utc) {
     if (window.moment) {
         var f = block.hash.format || "MMM Do, YYYY";
-        return moment(context).format(f);
+        return (utc === true) ? moment(context).utc().format(f) : moment(context).format(f);
     } else {
         //  moment plugin not available. return data as is.
         console.log("The moment.js plugin is not loaded.  Please make sure you have included moment.js on this page.");
--- a/composer.json
+++ b/composer.json
@@ -27,11 +27,17 @@
             "wikimedia/composer-merge-plugin": true
         }
     },
+    "repositories": [
+        {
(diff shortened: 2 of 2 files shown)


app/sprinkles/admin/assets/userfrosting/js/handlebars-helpers.js
```diff
@@ -60,10 +60,10 @@ Handlebars.registerHelper('calc', function (v1, operator, v2, options) {
  * usage: {{dateFormat creation_date format="MMMM YYYY"}}
  * @requires momentjs http://momentjs.com/
  */
-Handlebars.registerHelper('dateFormat', function(context, block) {
+Handlebars.registerHelper('dateFormat', function(context, block, utc) {
```
lcharette, line 63: Is UTC optional ?

lcharette (commented):
Do you have an exemple where this would be useful or an issue? Or is it an issue specific to your app ?

app/sprinkles/admin/assets/userfrosting/js/handlebars-helpers.js
phillmac: Yes it should be optional

phillmac:
I suppose it's a little bit specific, but I'm pretty sure it could be useful elsewhere too.
I have a series of GPS timestamps that are always in UTC regardless of any settings, time zone of the server, etc.
They're stored as a integer because that's the raw data from the gps module.
Being located in the  GMT+1000 most of the year and GMT+1100 the rest makes it a bit of a pain to do conversions on the server, and not even that is a sure thing. 
There might be some in GMT+0930 or GMT+0800 sometimes. 
So it sems to me that it might be easier to do the conversion browser side.
Not exactly sure if it's useful to anyone else, but at least one other person requested it sometime, that is, in the issue referenced above.

composer.json
```diff
@@ -27,11 +27,17 @@
...
+            "url": "https://github.com/phillmac/userfrosting-framework.git",
+            "type": "git"
+        }
+    ],
     "require": {
         "php": "^7.3 | ^8.0",
         "ext-gd": "*",
         "composer/installers": "^1.4.0",
-        "userfrosting/framework": "^4.6",
+        "phillmac/userfrosting-framework": "dev-master#4.6.3-sasl",
```
lcharette, line 40: FYI, I won't be able to merge with this

PR #1223, opened 2023-05-26
Better sprinkle discovery

avsdev-cw:
- Modified the sprinkle manager to first attempt to find a sprinkle's namespace from the "[sprinkle]/src/[sprinkle namespace].php" file, otherwise default to the old method of using "Str::studly" which can't handle consecutive capitals (e.g. UFTweaks)
- Removed all the instances where a sprinkle name was "guessed" and replaced them using calls to the sprinkle manager instead
- Updated tests to accept the usage of the sprinkle manager


lcharette:
Closing this, as it's not relevant with UF5 anymore

